PR #10, opened 2022-04-20
Search page / search results

jonnywyatt:
- against mocked integration tests only, as the actual API search endpoint doesn't exist yet


server/config.ts
```diff
@@ -33,6 +33,7 @@ export interface ApiConfig {
 
 export default {
   https: production,
+  applicationName: 'Recall Decisions',
```
jackmaloney, line 36: Might want to add the work "Make" to the application name

jackmaloney (approved):

server/config.ts
jonnywyatt: I asked Paul this yesterday and he definitely confirmed it's 'Recall Decisions', as in the prototype header

jonnywyatt: At least for the time being 🤷

PR #22, opened 2022-04-25
Case summary overview - person details block

jonnywyatt:
image


server/@types/make-recall-decision-api/models/Address.ts
```diff
@@ -0,0 +1,10 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+export type Address = {
+  line1: string
```
jackmaloney, line 6: do we want to make line1 and town optional as well potentially? Just thinking about the scenario with redacted cases, what would happen if this came back as empty?

PR #24, opened 2022-04-26
Run the E2E tests post-deployment

dazoakley:
This sets up the E2E tests to run after deployment to `dev` and `preprod` and block the promotion through the tiers if they fail.


.circleci/config.yml
```diff
@@ -133,6 +133,52 @@ jobs:
...
+          name: E2E Check - << parameters.environment >>
+          command: |
+            SPECS=$(circleci tests glob e2e_tests/integration/*.feature | circleci tests split --split-by=timings | tr "\n" "," | tr " " ",")
+            echo "Running feature spec(s): ${SPECS}"
+
+            npx cypress run \
+              --env USERNAME=${CYPRESS_USERNAME_<< parameters.environment >>},PASSWORD=${CYPRESS_PASSWORD_<< parameters.environment >>} \
+              --config-file e2e_tests/cypress.json \
+              --config baseUrl=https://make-recall-decision-<< parameters.environment >>.hmpps.service.justice.gov.uk \
+              --browser chrome \
```
jonnywyatt, line 164: Not a major issue at all, but if the image contains Edge browser, that's our main target browser (what probation use on their MoJ laptops) so might be worth using that?

jonnywyatt (dismissed):

.circleci/config.yml
dazoakley: Yeah can totally do that 👍

jonnywyatt (approved):

PR #35, opened 2022-04-29
MRD-28: Run the E2E tests "locally" on CI

dazoakley:
The way this works is pulling down both UI and API, spinning up the `docker-compose` stack in both repos (connected on the same virtual network), then running the E2E's over those services.

The downside to this is that it takes ~3 minutes to build and spin up the stack, then however long the E2E's take... Not a lot we can do about that though I'm afraid. 😞

### DIFF
--- a/scripts/start-services-for-e2e-tests.sh
+++ b/scripts/start-services-for-e2e-tests.sh
@@ -23,6 +23,8 @@ docker-compose build
 docker-compose up -d
 popd
 
+set +x
+
 function wait_for {
   printf "\n\nWaiting for %s to be ready.\n\n" "${2}"
   local TRIES=0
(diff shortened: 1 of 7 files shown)


scripts/start-services-for-e2e-tests.sh
```diff
@@ -0,0 +1,35 @@
...
+readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+readonly UI_NAME=make-recall-decision-ui
+readonly API_NAME=make-recall-decision-api
+readonly UI_DIR="${SCRIPT_DIR}/../../${UI_NAME}"
+readonly API_DIR="${SCRIPT_DIR}/../../${API_NAME}"
+
+pushd "${API_DIR}"
+printf "\n\nBuilding/starting API components...\n\n"
+docker-compose pull
```
jonnywyatt, line 14: in the UI readme the order is build -> pull not pull -> build
not sure if it matters tho?

jonnywyatt (approved):

scripts/start-services-for-e2e-tests.sh
dazoakley: Makes no difference

PR #39, opened 2022-05-03
Add script to run app locally; update readme

jonnywyatt:
also, point readiness probe to /health endpoint

### DIFF
--- a/scripts/start-services-no-ui.sh
+++ b/scripts/start-services-no-ui.sh
@@ -20,7 +20,7 @@ pushd "${UI_DIR}"
 printf "\n\nBuilding/starting UI components...\n\n"
 docker-compose pull
 docker-compose build
-docker-compose up -d
+docker-compose up --scale make-recall-decision-ui=0
 popd
 
 function wait_for {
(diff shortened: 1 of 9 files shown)


scripts/start-services-no-ui.sh
```diff
@@ -0,0 +1,38 @@
...
+function wait_for {
+  printf "\n\nWaiting for %s to be ready.\n\n" "${2}"
+  until curl -s --fail "${1}"; do
+    printf "."
+    sleep 2
+  done
+}
+
+wait_for "http://localhost:9090/auth/health/ping" "hmpps-auth"
+wait_for "http://localhost:3000/ping" "${UI_NAME}"
```
dazoakley, line 35: Thia will fail unless you start up UI in a separate terminal session - might be best to remove this check and add something to the closing output to say "and now start up UI by doing ..."

jonnywyatt: removed

dazoakley (approved):

dazoakley (approved):

PR #53, opened 2022-05-11
Run docker compose in API now we have dependencies

dazoakley:
ref https://github.com/ministryofjustice/make-recall-decision-api/pull/52

### DIFF
--- a/scripts/start-services-no-ui.sh
+++ b/scripts/start-services-no-ui.sh
@@ -62,8 +62,7 @@ fi
 
 pushd "${API_DIR}"
 printf "\n\nBuilding/starting API components...\n\n"
-# TODO: uncomment the below when we have any API dependencies to run
-#docker-compose up -d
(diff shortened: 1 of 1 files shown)


scripts/start-services-no-ui.sh
```diff
@@ -62,8 +62,7 @@ fi
 
 pushd "${API_DIR}"
 printf "\n\nBuilding/starting API components...\n\n"
-# TODO: uncomment the below when we have any API dependencies to run
-#docker-compose up -d
+docker-compose up -d
```
jonnywyatt, line 65: will this run the API itself as a container? In which case should we remove the gradle command below?

dazoakley: 🤦 yep - let me fix

jonnywyatt (approved):

jonnywyatt (approved):

PR #102, opened 2022-05-24
MRD-148 - date filter enhancements

jonnywyatt:
- prefix page title with number of contact results, for screen readers
- set result count as an H2, for easier location by screen readers
- more specific date validation - indicate which part of the date is invalid, and link to that input from the error summary
- add date filter URLs to the accessibility tests

### DIFF
--- a/server/utils/dates/convert.ts
+++ b/server/utils/dates/convert.ts
@@ -17,7 +17,31 @@ const filterPartsForEmptyStrings = (parts: unknown[]): DatePartNames[] =>
   parts.map(({ name, value }) => (value === '' ? name : undefined)).filter(Boolean)
 
 const filterPartsForMinimumLength = (parts: unknown[]): DatePartNames[] =>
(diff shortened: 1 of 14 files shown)


server/utils/dates/convert.ts
```diff
@@ -17,7 +17,31 @@ const filterPartsForEmptyStrings = (parts: unknown[]): DatePartNames[] =>
...
+
+const filterPartsForValueRange = (parts: unknown[]): DatePartNames[] =>
+  parts
+    .map(({ name, numberValue, minValue, maxValue }) =>
+      numberValue < minValue || numberValue > maxValue ? name : undefined
+    )
+    .filter(Boolean)
+
+const convertDatePartsToNumbers = (parts: DateTimePart[]) => {
+  return parts.map(part => {
```
jackmaloney, line 30: Do we have unit tests to cover these new functions?

bills7281 (approved):

jackmaloney (approved):

PR #322, opened 2022-09-22
MRD-626 - Reasons for no recall

jonnywyatt:


server/textStrings/en.ts
```diff
@@ -95,5 +100,6 @@ export const strings = {
     confirmationPartA: 'Part A created',
     taskListNoRecall: 'Create a Decision not to Recall letter',
     whyConsideredRecall: 'Why you considered recall',
+    reasonsForNoRecall: 'Why you think the person should not be recalled?',
```
jackmaloney, line 103: How come this error doesn't have the persons name in it?

jonnywyatt: Page titles are sent to Google Analytics and shouldn't contain PII
There is some duplication in these two objects though, which is on my list to refactor.

jackmaloney (approved):

PR #401, opened 2022-11-03
MRD-855 - Send audit events for rec flow page views

jonnywyatt:


server/services/auditService.ts
```diff
@@ -62,6 +62,31 @@ export class AuditService {
...
+    recommendationId: string
+    pageUrlSlug: string
+    crn: string
+    username: string
+    logErrors: boolean
+  }) {
+    return this.sendAuditMessage({
+      action: 'VIEWED_RECOMMENDATION_PAGE',
+      who: username,
+      details: {
```
jackmaloney, line 81: As an FYI, this 'details' section is the part we can't see in App Insights.

jackmaloney (approved):

PR #406, opened 2022-11-03
MRD-851 Fire custom event when make a recommendation button clicked

jackmaloney:


server/controllers/recommendations/createRecommendation.ts
```diff
@@ -2,10 +2,15 @@ import { Request, Response } from 'express'
...
 import { routeUrls } from '../../routes/routeUrls'
+import { trackEvent } from '../../monitoring/azureAppInsights'
+import { EVENTS } from '../../utils/constants'
+import logger from '../../../logger'
 
 export const createRecommendationController = async (req: Request, res: Response): Promise<Response | void> => {
   const normalizedCrn = validateCrn(req.body.crn)
   try {
+    logger.info(`About to track the MRD_RECOMMENDATION_STARTED event to app insights`)
+    trackEvent(EVENTS.MRD_RECOMMENDATION_STARTED, req)
```
jonnywyatt, line 13: I'd move this after line 15 (the redirect). That way, if it throws, the redirect will still happen.

bills7281 (approved):

server/utils/constants.js
```diff
@@ -0,0 +1,7 @@
+const EVENTS = {
```
jonnywyatt, line 1: Maybe one for later but I'd consider using a typescript enum here instead of an object literal. The benefit of that would be, if `trackEvent` was called with anything other than a valid event name, the code won't even compile so the bug will be caught immediately.

server/monitoring/azureAppInsights.ts
```diff
@@ -27,9 +30,23 @@ export function initialiseAppInsights(): void {
...
+
+export const trackEvent = (eventName: string, req: Request) => {
+  const requestBody = req.body
+
+  if (appInsightsClient && eventName) {
+    const eventProperties = {
+      crn: requestBody?.crn,
+      author: req.user?.username,
+    }
+    appInsightsClient.trackEvent({ name: eventName, properties: { ...eventProperties } })
```
jonnywyatt, line 50: No need to spread `eventProperties` here, just assign it to properties. Or even shorter, call the variable `properties` then the assignment is `... , properties }`

server/utils/constants.js
```diff
@@ -0,0 +1,7 @@
...
+  MRD_RECOMMENDATION_STARTED: 'mrdRecommendationStarted',
+}
+
+module.exports = Object.freeze({
```
jonnywyatt, line 5: Named exports are more helpful as the IDE will suggest the names as you start typing

jonnywyatt: It's consistent to use a .ts file extension here

server/monitoring/azureAppInsights.ts
```diff
@@ -27,9 +30,23 @@ export function initialiseAppInsights(): void {
...
   return null
 }
+
+export const trackEvent = (eventName: string, req: Request) => {
+  const requestBody = req.body
+
+  if (appInsightsClient && eventName) {
+    const eventProperties = {
+      crn: requestBody?.crn,
+      author: req.user?.username,
```
jonnywyatt, line 48: Is 'author' the best name here? How about just `userName`

jonnywyatt (changes requested):

server/monitoring/azureAppInsights.ts
jackmaloney: yeh was just copying the pattern that ARN used, can change

server/utils/constants.js
jackmaloney: Fixed

server/utils/constants.js
jackmaloney: Fixed

server/monitoring/azureAppInsights.ts
jonnywyatt: ah ok I guess they're calling it that as a report is being generated
🤷

jonnywyatt (approved):

PR #595, opened 2023-03-14
MRD-1132 - changes to reasonsForNoRecall page

robinandrewdevilliers:


DannyMClegg (approved):
In the future, I think we need to talk about any changes to scripts as a team. Also, any script changes should be a separate ticket.

PR #661, opened 2023-05-03
MRD-1313 - Tests for the story and some minor enhancement

johnson-moj:
- Checks the Presence of 'Action Required' Banner softly (meaning wont fail if found to be wrong immediately but will wait for all subsequent steps to finish before reporting the failure)
- Adds the SPO countersignature flow for non-indeterminate, non-extended, non-emergency recall
- Changed PartA document assertions to soft assertion so that all assertions are carried out
- Fixed the test data for dev
- Replaced 'defineStep' with 'When'


robinandrewdevilliers (approved):
looks good

PR #712, opened 2023-06-14
MRD-1450 - countersigning variant of spo rationale page

robinandrewdevilliers:


e2e_tests/integration/rationale.feature
```diff
@@ -180,7 +180,7 @@ Feature: Rationale for a Recall decision
     # MRD-1262: AC0/AC1
     And PO requests an SPO to countersign
     And SPO has visited the countersigning link
-    And SPO has recorded rationale with <SPODecision> decision
```
johnson-moj, line 183: This step was written to cater to both RECALL & NO RECALL decisions and before SPO has countersigned, if the intention of this step is to be called SPO countersignature, then there's another step called, which need to be changed to only allow RECALL

```ghekin
Then SPO is able to record rationale with RECALL decision
```
```javascript
Then('SPO is able to record rationale with {managersDecision} decision', function (decision: string) {
  recordSpoDecision.call(this, decision)
})
```

e2e_tests/integration/stepDefinitions/index.ts
```diff
@@ -66,16 +66,17 @@ export const deleteOpenRecommendation = () => {
...
+      for (let i = 0; i < length; i += 1) {
+        cy.getRowValuesFromTable({ tableCaption: 'Recommendations', rowSelector: 'tr[data-qa]:first-child' }).then(
+          // eslint-disable-next-line no-loop-func
+          () => {
+            cy.get('[data-qa] [data-qa="delete-recommendation"]').first().click()
           }
-        }
-      )
+        )
+      }
```
johnson-moj, line 79: We can check for presence of `Make a recommendation` button and break if found:

```javascript
if ($body.find('#main-content a:contains("Make a recommendation")').length) break
```

PR #732, opened 2023-06-26
MRD-1488 write end to end test for the DNTR journey

mayank-s18:
Write an e2e test for following scenario  - Decision Not To Recall Letter details are correct when PO and SPO records rationale during review and countersign

### DIFF
--- a/e2e_tests/utils/index.ts
+++ b/e2e_tests/utils/index.ts
@@ -32,3 +32,51 @@ export const formatObjectDateToLongFormat = (objectDate: Record<string, number |
 export const changeDateFromLongFormatToShort = (dateToConvert: string) => {
   return DateTime.fromFormat(dateToConvert, 'dd MMMM yyyy').toFormat('dd/MM/yyyy')
 }
+
+export const formatDateToDNTRLetterFormat = objectDate => {
+  const d = objectDate
+  const months = [
+    'January',
+    'February',
+    'March',
+    'April',
+    'May',
+    'June',
+    'July',
+    'August',
+    'September',
+    'October',
+    'November',
+    'December',
+  ]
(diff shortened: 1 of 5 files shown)


e2e_tests/utils/index.ts
```diff
@@ -32,3 +32,51 @@ export const formatObjectDateToLongFormat = (objectDate: Record<string, number |
...
+      case 2:
+        return 'nd'
+      case 3:
+        return 'rd'
+      default:
+        return 'th'
+    }
+  }
+  return `${weekDay} ${date}${nth(date)} ${month} ${year} at ${hr}:${min}${ampm}`
+}
```
johnson-moj, line 82: This can be achieved using `luxon` as well, did you try this, will save a few lines of code. Its not mandatory to accept this, it's just a suggestion :)

e.g. changes `Thu Jun 29 2023 at 05:22:47 GMT+0100 (British Summer Time)` to `Thursday 29th June 2023 at 5:22AM`

```suggestion
export const formatDateToDNTRLetterFormat = (objectDate: Date) => {
  const nth = function (rawDate) {
    if (rawDate > 3 && rawDate < 21) return 'th'
    switch (rawDate % 10) {
      case 1:
        return 'st'
      case 2:
        return 'nd'
      case 3:
        return 'rd'
      default:
        return 'th'
    }
  }
  cy.log(`objectDate--> ${objectDate.toDateString()} at ${objectDate.toTimeString()}`)
  return DateTime.fromObject({
    year: objectDate.getFullYear(),
    month: objectDate.getMonth() + 1,
    day: objectDate.getDate(),
    weekday: objectDate.getDay(),
    hour: objectDate.getHours(),
    minute: objectDate.getMinutes(),
    second: objectDate.getSeconds(),
  }).toFormat(`EEEE d'${nth(objectDate.getDate())}' MMMM yyyy 'at' h':'mma`)
}
```

PR #754, opened 2023-07-19
MRD-1566 - present cvl data on CVL tab

robinandrewdevilliers:


bills7281 (approved):

bills7281:
Hi Robin, take this as a non-blocking comment as I could be wrong.
When I test this with stubs , I get the not on licence banner appearing for all cases.
It looks to me like the [convictionLicenceBanner.njk template](https://github.com/ministryofjustice/make-recall-decision-ui/blob/816811c2390f3ce49f17428332d5eb0761470e97/server/views/partials/notificationBanners/convictionsLicencesBanner.njk#L22) needs an update on line 22 to consider the `hasAllConvictionsReleasedOnLicence `flag on the V2 response.

I'm not sure if I've implemented it correctly, but when I add the change below it works.

image

robinandrewdevilliers:
> Hi Robin, take this as a non-blocking comment as I could be wrong. When I test this with stubs , I get the 'not on licence' banner appearing for all cases. It looks to me like the [convictionLicenceBanner.njk template](https://github.com/ministryofjustice/make-recall-decision-ui/blob/816811c2390f3ce49f17428332d5eb0761470e97/server/views/partials/notificationBanners/convictionsLicencesBanner.njk#L22) needs an update on line 22 to consider the `hasAllConvictionsReleasedOnLicence `flag on the V2 response.
> 
> I'm not sure if I've implemented it correctly, but when I add the change below it works as expected.
> 
> image

tks Bill - yes - spotted that myself a few minutes ago - should have seen your message.

PR #791, opened 2023-08-10
MRD-1700 for e2e tests explicitly set timezone so it does not cater f…

mayank-s18:
…or UTC by default


mayank-s18:
failing on CI

PR #814, opened 2023-08-21
MRD-1714 - steal thread for contact information section of task list

robinandrewdevilliers:


integration_tests/integration/recommendationTaskList.spec.ts
```diff
@@ -235,6 +235,18 @@ context('Recommendation - task list', () => {
...
+  it('task list - review and send', () => {
+    cy.task('getRecommendation', {
+      statusCode: 200,
+      response: { ...recommendationResponse, isIndeterminateSentence: true, custodyStatus: { selected: 'NO' } },
+    })
+    cy.task('getStatuses', { statusCode: 200, response: [] })
+    cy.visit(`${routeUrls.recommendations}/${recommendationId}/task-list?flagReviewAndSend=1`)
+
+    cy.getElement('Who completed this Part A?').should('exist')
+    cy.getElement('Where should the revocation order by sent?').should('exist')
```
DarrenW-HMPPS, line 247: Minor typo here - should be "order *be* sent"

server/middleware/featureFlags.ts
```diff
@@ -4,6 +4,11 @@ import { FeatureFlagDefault } from '../@types/featureFlags'
 import { isPreprodOrProd } from '../utils/utils'
 
 export const featureFlagsDefaults = {
+  flagReviewAndSend: {
+    label: 'Enable Review And Send',
```
DarrenW-HMPPS, line 8: Should this refer to "Probation Admin" rather than "Review and Send"?

server/views/partials/taskList/reviewAndSend.njk
```diff
@@ -0,0 +1,25 @@
+<li>
+    <h2 class="moj-task-list__section" id='contact-information'>
+        Contact Information
+    </h2>
+    <ul class="moj-task-list__items">
+        <li class="moj-task-list__item">
+            <span class="moj-task-list__task-name">
+                <a aria-describedby='task-list-status-whoCompletedPartA' href="{{ urlInfo.basePath }}xyz">{{ pageHeadings.whoCompletedPartA }}</a>
```
DarrenW-HMPPS, line 8: Is xyz a placeholder? Can we use something better than that?

DarrenW-HMPPS (changes requested):

server/controllers/recommendations/helpers/taskCompleteness.ts
```diff
@@ -107,6 +107,8 @@ export const taskCompleteness = (recommendation: RecommendationResponse, _featur
       recommendation.custodyStatus?.selected !== 'NO' || hasValue(recommendation.isMainAddressWherePersonCanBeFound),
   }
 
+  // const flagReviewAndSend = _featureFlags.flagReviewAndSend
```
DarrenW-HMPPS, line 110: Should this be flagProbationAdmin?

robinandrewdevilliers: Good spot - it's commented out code anyway.  Can we relax this level of correctness please?  That would be fixed up pretty soon, when the real code goes in.  And re committing and running through CI will cost me about 40 minutes.

DarrenW-HMPPS: Yes, CI takes too long and is quite painful at the moment. It ought to be 10 minutes at the most, but it's been allowed to extend way too long.  We really should pay some attention to it.
As regards the level of correctness, if we are going to leave in commented out code to ease introduction in the future, then it should be correct, or it doesn't make it easier in the future.  I would prefer no commented code at all, but have let that pass.

DarrenW-HMPPS (approved):

PR #815, opened 2023-08-22
mrd-1491_banner-for-maintenance-outage: added maintenance page

bills7281:


server/views/components/question/macro.njk
```diff
@@ -0,0 +1,3 @@
+{% macro renderQuestion(question, errors, options, bodyAnswer) %}
```
DarrenW-HMPPS, line 1: Have these 2 files in the question folder been included by accident?

DarrenW-HMPPS (changes requested):

server/views/components/question/macro.njk
bills7281: I took them from the maintenance page example Andrew provided - https://github.com/ministryofjustice/hmpps-risk-assessment-ui/pull/651/files it uses the 2 extra macros. They aren't present in our project so that's why I added them here.

DarrenW-HMPPS (approved):

PR #825, opened 2023-08-30
Streamline the e2e local tests to improve the pipeline execution time

mayank-s18:


.circleci/config.yml
```diff
@@ -322,7 +322,7 @@ jobs:
             
             shopt -s globstar
             
-            SPECS=$(circleci tests glob "e2e_tests/features/**/*.feature" | circleci tests split --split-by=timings --show-counts)
+            SPECS=$(circleci tests glob "e2e_tests/features/*.feature" | circleci tests split --split-by=timings --show-counts)
```
DarrenW-HMPPS, line 325: Doesn't this mean that it won't run the tests in the features/e2e folder? Or any other folder that we choose to create in the future?

DarrenW-HMPPS (changes requested):

.circleci/config.yml
mayank-s18: The tests under features/e2e folder are currently skipped for e2e_local_test job. But even to skip these tests cypress spends a few mins on the pipeline to execute it and skip it. By excluding the features/e2e folder we should save those few mins cypress takes, hopefully.

DarrenW-HMPPS: I see.  That seems to be somewhat hidden and not obvious that it is happening.  Can we make it more explicit with appropriate folder hierarchy and naming?

mayank-s18: updated as requested..

.circleci/config.yml
```diff
@@ -322,7 +322,7 @@ jobs:
             
             shopt -s globstar
             
-            SPECS=$(circleci tests glob "e2e_tests/features/**/*.feature" | circleci tests split --split-by=timings --show-counts)
+            SPECS=$(circleci tests glob "e2e_tests/features/e2eLocalTests.feature" | circleci tests split --split-by=timings --show-counts)
```
DarrenW-HMPPS, line 325: I think this might need to be e2e_tests/features/e2eLocalTests/*.feature

And similar for the one above

DarrenW-HMPPS (approved):

PR #846, opened 2023-09-06
Mrd 1491 maintenance banner

bills7281:


robinandrewdevilliers (changes requested):
It looks like the notification stuff you are adding to res.locals is a cross cutting concern.  There is an appropriate way to handle this stuff using express js.

Have a look at recommendations.ts. There are some builders in there that wire up the controllers - add one and it will affect all controllers.  Ping me and I can walk  you through it.

robinandrewdevilliers:
:) Sorry Bill...   Please add your own middleware, and remember a unit test please.

robinandrewdevilliers:
You could also add your middleware at the root router - and then you won't have to add code to the caseSummaryController as well.

robinandrewdevilliers (approved):

PR #852, opened 2023-09-08
mrd-1491_maintenance-banner-correction: set fallback value for showin…

bills7281:
…g the banner to false


server/controllers/personSearch/personSearchByCRN.ts
```diff
@@ -4,7 +4,7 @@ import config from '../../config'
 export const personSearchByCRN = async (req: Request, res: Response): Promise<Response | void> => {
   res.locals.notification = {
     ...config.notification,
-    isVisible: config.notification.body && config.notification.active,
+    isVisible: Boolean(config.notification.active === 'true'.toLowerCase()) && Boolean(config.notification.body),
```
robinandrewdevilliers, line 7: 'true'.toLowerCase() is redundant.

You've just explicitly made it lowercase.

The enclosing Boolean is redundant as well.  

If you want you can use the ! or !! as a boolean converter.  But what you've done here is fine.

ie.  !!config.notification.body

robinandrewdevilliers (changes requested):

server/controllers/personSearch/personSearchByCRN.ts
bills7281: Thanks! All resolved!

DarrenW-HMPPS (approved):

PR #894, opened 2023-09-29
Mrd 1202 1

robinandrewdevilliers:


integration_tests/integration/excludedRestricted.spec.ts
```diff
@@ -11,15 +9,6 @@ context('Excluded and restricted cases', () => {
   })
 
   context('Excluded', () => {
-    it('search results page', () => {
```
DarrenW-HMPPS, line 14: Do we not need some sort of equivalent of this test for search by name?

integration_tests/integration/personSearch.spec.ts
```diff
@@ -122,7 +78,7 @@ context('Search for a person', () => {
   })
 
   it('can search for a person on probation by name with search by name flag', () => {
```
DarrenW-HMPPS, line 80: Need to drop the "with search by name flag" from the test name

integration_tests/integration/personSearch.spec.ts
```diff
@@ -80,7 +36,7 @@ context('Search for a person', () => {
 
   it('can search for a person on probation by CRN with search by name flag', () => {
```
DarrenW-HMPPS, line 37: Need to drop the "with search by name flag" from the test name

bills7281 (approved):

DarrenW-HMPPS (changes requested):

integration_tests/integration/excludedRestricted.spec.ts
robinandrewdevilliers: no - since the endpoint now returns an array - it can't complain about excluded cases anymore.  We encounter that error when we arrive on the case overview.

integration_tests/integration/personSearch.spec.ts
robinandrewdevilliers: resolved

integration_tests/integration/personSearch.spec.ts
robinandrewdevilliers: resolved

DarrenW-HMPPS (approved):

PR #943, opened 2023-10-21
mrd-1975_ppcs-logins: PPCS role added with first screen of journey

bills7281:
image


server/controllers/startPage/startPage.test.ts
```diff
@@ -1,14 +1,50 @@
...
+}
+function createResWithToken(
+  { authorities }: { authorities: string[] },
+  { currentPageId }: { currentPageId: string }
+): Response {
+  return {
+    token: createToken(authorities),
+    locals: {
+      flags: {
+        flagProbationAdmin: 'true',
```
DarrenW-HMPPS, line 26: This work should be separate to the probationAdmin flag.  We can perhaps just delete this, but it might mean it's been linked where it shouldn't be.

server/controllers/startPage/startPage.ts
```diff
@@ -1,12 +1,25 @@
 import { Request, Response } from 'express'
+import jwtDecode from 'jwt-decode'
 import config from '../../config'
 import { isBannerDisplayDateRangeValid } from '../../utils/utils'
 
+export enum HMPPS_AUTH_ROLE {
+  PPCS = 'ROLE_MARD_PPCS',
+}
```
DarrenW-HMPPS, line 8: The existing roles we have are:
ROLE_MAKE_RECALL_DECISION
ROLE_MAKE_RECALL_DECISION_SPO

It would be much better if this new one conformed to the same pattern. i.e.
ROLE_MAKE_RECALL_DECISION_PPCS

DarrenW-HMPPS (changes requested):

server/controllers/startPage/startPage.ts
bills7281: Eek! The HMPPS-Auth team, and Sam have already added that name to dev and to the hmpps-auth service for local use!
Er... shall I create new PRs/requests to change that? Or are we ok to keep it as it is?

server/controllers/startPage/startPage.test.ts
bills7281: It isn't linked to probationAdmin, I added the flag there accidentally as a cut/paste job :O

server/controllers/startPage/startPage.ts
DarrenW-HMPPS: I think it would be better to change it.  A little bit of pain now will avoid possibly frequent confusion/frustration/mistakes in the future.

bills7281: Sounds good!

server/controllers/startPage/startPage.test.ts
bills7281: Done :D

server/controllers/startPage/startPage.ts
bills7281: Done 👍

DarrenW-HMPPS (approved):

PR #944, opened 2023-10-23
Mrd 1940

robinandrewdevilliers:


server/controllers/recommendation/reviewPractitionersConcernsController.ts
```diff
@@ -7,15 +7,15 @@ import { AdditionalLicenceConditionOption } from '../../@types/make-recall-decis
 function extractStandardLicenceConditions(recommendation: RecommendationDecorated): Array<string> {
   if (recommendation.licenceConditionsBreached && recommendation.licenceConditionsBreached.standardLicenceConditions) {
     const { selected, allOptions } = recommendation.licenceConditionsBreached.standardLicenceConditions
-    return selected.map((s: string) => allOptions.find((o: Record<string, string>) => o.value === s).text)
+    return selected.map((s: string) => allOptions.find((o: Record<string, string>) => o.value === s)?.text)
```
DarrenW-HMPPS, line 10: I think we're going to end up returning an Array<string> that potentially contains elements that are undefined.
allOptions.find returns undefined if no matches are found.  It would be better if made sure the returned array only contained valid strings.
Is this the change that you mentioned in Slack this morning to address the 500 errors? It feels to me like this code should be able to assume that `selected` is a subset of `allOptions`.  Is there something else that is wrong (i.e. the data), rather than the issue being in this code?

server/controllers/recommendation/reviewPractitionersConcernsController.ts
```diff
@@ -7,15 +7,15 @@ import { AdditionalLicenceConditionOption } from '../../@types/make-recall-decis
...
   }
 
   if (
     recommendation.cvlLicenceConditionsBreached &&
     recommendation.cvlLicenceConditionsBreached.standardLicenceConditions
   ) {
     const { selected, allOptions } = recommendation.cvlLicenceConditionsBreached.standardLicenceConditions
-    return selected.map((s: string) => allOptions.find((o: Record<string, string>) => o.code === s).text)
+    return selected.map((s: string) => allOptions.find((o: Record<string, string>) => o.code === s)?.text)
```
DarrenW-HMPPS, line 18: Same here of course.

server/controllers/recommendation/reviewPractitionersConcernsController.ts
```diff
@@ -44,7 +44,7 @@ function extractAdditionalLicenceConditions(recommendation: RecommendationDecora
     const { selected, allOptions } = recommendation.cvlLicenceConditionsBreached.additionalLicenceConditions
     return selected.map((s: string) => {
       return {
-        title: allOptions.find((o: Record<string, string>) => o.code === s).text,
+        title: allOptions.find((o: Record<string, string>) => o.code === s)?.text,
```
DarrenW-HMPPS, line 47: And similar but different here.  Empty string is probably better than undefined.

DarrenW-HMPPS (changes requested):

server/controllers/recommendation/reviewPractitionersConcernsController.ts
robinandrewdevilliers: yeah ok, I'll back that change out.

DarrenW-HMPPS (approved):

PR #952, opened 2023-10-26
mrd-1970_ppcs-user-search-screen: search screen for PPCS users journey

bills7281:


server/views/pages/paginatedPersonSearchResults.njk
```diff
@@ -24,56 +23,73 @@
...
-        <p class="govuk-body-l govuk-!-margin-bottom-3">
-            <span class='govuk-!-padding-right-2 govuk-!-font-weight-bold'>Case reference number (CRN):</span>
-             <span>{{ crn }}</span>
-        </p>
-        <a href="/search-by-crn" class="govuk-body-l govuk-link">Change <span class="govuk-visually-hidden">search term</span></a>
+            <p class="govuk-body-l govuk-!-margin-bottom-3">
+                <span class='govuk-!-padding-right-2 govuk-!-font-weight-bold'>Case reference number (CRN):</span>
+                <span>{{ crn }}</span>
+            </p>
+            {% if user.hasPpcsRole %}
```
DarrenW-HMPPS, line 30: What about passing this in from the earlier page as a query parameter saying "the search page I have come from". That means we don't have to revisit this page if we ever create a new search page.  It also means that we don't need to get involved in checking role membership.

server/views/pages/paginatedPersonSearchResults.njk
```diff
@@ -24,56 +23,73 @@
...
-                </td>
-                <td class="govuk-table__cell">{{ person.crn }}</td>
-                <td class="govuk-table__cell">{{ formatDateTimeFromIsoString({isoDate: person.dateOfBirth}) }}</td>
-            </tr>
-        {% endfor %}
-        </tbody>
-    </table>
-    {{ paginate({
+
+                            {% if user.hasPpcsRole %}
```
DarrenW-HMPPS, line 72: Would benefit from an auto format I think.

DarrenW-HMPPS: I think we should probably be honouring the Delius restrictions for PPCS users rather than overriding them.  We believe that currently PPCS can see everything, but we should expect/allow that to be specified in Delius.

server/views/pages/personSearchResults.njk
```diff
@@ -1,44 +1,56 @@
 {% extends "../partials/layout.njk" %}
 
-
```
DarrenW-HMPPS, line 3: Is this also old code we don't need? Do we always paginate now?

DarrenW-HMPPS (changes requested):

server/controllers/personSearch/personSearchResultsByName.ts
```diff
@@ -17,7 +17,7 @@ export const personSearchResultsByName = async (req: Request, res: Response) =>
     page: req.query.page as string,
   }
   const { user, flags } = res.locals
-
+  res.locals.user.hasPpcsRole = user.roles.includes('ROLE_MAKE_RECALL_DECISION_PPCS')
```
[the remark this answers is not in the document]
bills7281: I was going to do that but then realised that we'd need to update all the unit tests (probably 100s), so I opted out :O 
I can still do it if you like, I'll incline to my reviewer's choice here! What do you think?

server/views/pages/findaRecallRequestByCRN.njk
```diff
@@ -0,0 +1,45 @@
+{% extends "../partials/layout.njk" %}
```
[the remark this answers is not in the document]
bills7281: Done :D !!

server/views/pages/findaRecallRequestByCRN.njk
```diff
@@ -0,0 +1,45 @@
...
+
+
+{% set pageTitle = makePageTitle({ pageHeading: "Find a Recall Request" }) %}
```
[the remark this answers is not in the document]
bills7281: Removed!

server/views/pages/paginatedPersonSearchResults.njk
bills7281: Nice idea! Done!

server/views/pages/paginatedPersonSearchResults.njk
bills7281: Done!

server/views/pages/personSearchResults.njk
bills7281: Nice catch! Removed!

server/controllers/personSearch/personSearchResultsByName.ts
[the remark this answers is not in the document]
bills7281: Good idea! Done :)

DarrenW-HMPPS (approved):

PR #964, opened 2023-11-02
mrd-1972_ppud-search: dummy ppud search screen

bills7281:


server/routes/index.ts
```diff
@@ -35,6 +36,7 @@ export default function routes(router: Router): Router {
   get(routeUrls.searchByName, personSearchByName)
   get(routeUrls.searchResultsByCRN, personSearchResults)
   get(routeUrls.searchResultsByName, personSearchResultsByName)
+  get(routeUrls.searchInPpud, personSearchInPpud)
```
robinandrewdevilliers, line 39: I think that by  the time we are searching in ppud, that we are in the context of a recommendation doc, since we have found the rec doc in the previous pages.  
So all urls should be /recommendation/x/something.
And this controller should be bound into the recommendation routes, using the routebuilder.

Also please follow the get, post convention going forward.

You won't be able to do this ^ until we fix the route builder as it is. So perhaps push this PR in anyway.

server/views/pages/personSearchInPpudResults.njk
```diff
@@ -0,0 +1,51 @@
+{% extends "../partials/layout.njk" %}
+
```
robinandrewdevilliers, line 2: This page will need to move into the recommendations dir when we wire it correctly.

robinandrewdevilliers (approved):

PR #982, opened 2023-11-15
mrd-2089_change-the-point-when-a-recommendation-is-closed: added new …

bills7281:
…PPCS statuses to recall and no recall paths in backwards compatible manner


server/controllers/recommendations/createAndDownloadDocument.ts
```diff
@@ -62,13 +62,17 @@ export const createAndDownloadDocument =
...
+            activate.push(STATUSES.SENT_TO_PPCS)
+          }
         }
       }
     }
     if (documentType === 'NO_RECALL_LETTER') {
       const isCompleted = statuses.find(status => status.name === STATUSES.COMPLETED)
       if (!isCompleted) {
         activate.push(STATUSES.COMPLETED)
+        activate.push(STATUSES.DNTR_DOWNLOADED)
```
DarrenW-HMPPS, line 75: Should this be outside of the if?  Or perhaps even have its own if - if no DNTR_DOWNLOADED status, then add one?

bills7281: I've moved it to it's own if block to follow the existing pattern.

DarrenW-HMPPS (approved):

PR #1004, opened 2023-12-07
mrd-2114_clean-up-closed-statuses: references to closed statuses removed

bills7281:


server/controllers/recommendations/createAndDownloadDocument.ts
```diff
@@ -61,7 +61,6 @@ export const createAndDownloadDocument =
 
         const isSpoRecordedRationale = statuses.find(status => status.name === STATUSES.SPO_RECORDED_RATIONALE)
         if (!isSpo && isSpoRecordedRationale) {
-          activate.push(STATUSES.CLOSED)
```
DarrenW-HMPPS, line 64: Have we missed something here?  We would have set the closed state for PART As and DNTRs before this change, but we now set SENT_TO_PPCS just for Part As.  Is that right?  DNTRs aren't sent to PPCS, but is there something we still need to do here for DNTRs?

server/controllers/recommendations/replaceCurrentRecommendationController.ts
```diff
@@ -15,14 +15,14 @@ async function get(req: Request, res: Response) {
     })
   ).filter(status => status.active)
 
-  const isClosed = statuses.find(status => status.name === STATUSES.CLOSED)
+  const isClosed = statuses.find(status => status.name === STATUSES.SENT_TO_PPCS)
   const isPPDocumentCreated = statuses.find(status => status.name === STATUSES.PP_DOCUMENT_CREATED)
 
   if (!isClosed && isPPDocumentCreated) {
```
DarrenW-HMPPS, line 21: I think I've got a similar question here, but in reverse.  Looks like this will set SENT_TO_PPCS for DNTR and Recall.
Or does PP_DOCUMENT_CREATED mean "Part A created" and not "Part A or DNTR created"?

DarrenW-HMPPS (changes requested):

server/controllers/recommendations/createAndDownloadDocument.ts
bills7281: Yes DNTRs are now set to REC_CLOSED, see line 77

PR #1008, opened 2023-12-12
MRD-2049 - book to ppud

robinandrewdevilliers:


server/controllers/recommendation/checkBookingDetailsController.test.ts
```diff
@@ -99,6 +99,19 @@ describe('get', () => {
           gender: 'Male',
           locationDescription: 'Graceland',
         },
+        bookRecallToPpud: {
+          decisionDateTime: '2023-11-13T09:49:31.361Z',
```
DarrenW-HMPPS, line 103: PPUD Automation is expecting a local datetime for this, not a datetime with a timezone.

robinandrewdevilliers: good spot

DarrenW-HMPPS (approved):

PR #1016, opened 2023-12-20
no_jira-npe-fix-on-risk-dates-extended-sentence-fix: correction to ex…

bills7281:
…tended sentence logic and minor change to wording on screen

### DIFF
--- a/server/controllers/recommendation/selectIndexOffenceController.ts
+++ b/server/controllers/recommendation/selectIndexOffenceController.ts
@@ -5,7 +5,7 @@ import { NamedFormError } from '../../@types/pagesForms'
 import { RecommendationResponse } from '../../@types/make-recall-decision-api'
 import { makeErrorObject } from '../../utils/errors'
 import { strings } from '../../textStrings/en'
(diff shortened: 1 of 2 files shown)


server/utils/utils.ts
```diff
@@ -185,6 +185,8 @@ export function isMandatoryTextValue(val: unknown): boolean {
 
 export const isEmptyStringOrWhitespace = (val: string | string[]) => !val || !(val as string).trim()
 
+export const isNullEmptyStringOrWhitespace = (text: string) => text == null || text.match(/^\s*$/) !== null
+
```
DarrenW-HMPPS, line 189: I don't think this is any different in functionality to the isEmptyStringOrWhitespace on the line above.

DarrenW-HMPPS (changes requested):

server/utils/utils.ts
bills7281: ah ok! Good catch! Corrected!

DarrenW-HMPPS (approved):

PR #1033, opened 2024-01-09
MRD-2238 - edit names

robinandrewdevilliers:


server/textStrings/en.ts
```diff
@@ -14,6 +14,7 @@ export const strings: Record<string, Record<string, string>> = {
     missingSpoNoRecallRationale: 'You must explain your decision',
     missingLastName: 'Enter a last name',
     missingFirstName: 'Enter a first name',
+    missingFirstNames: 'Enter a first name(s)',
```
DarrenW-HMPPS, line 17: Think this should be "Enter first name(s)" rather than "Enter a first name(s)", as per Figma.

DarrenW-HMPPS (dismissed):

server/textStrings/en.ts
robinandrewdevilliers: tks

PR #1047, opened 2024-01-13
mrd-2246_edit-date-of-birth: dob editable from check booking details …

bills7281:
…screen

### DIFF
--- a/server/controllers/recommendation/editDateOfBirthController.test.ts
+++ b/server/controllers/recommendation/editDateOfBirthController.test.ts
@@ -16,11 +16,8 @@ describe('get', () => {
     const res = mockRes({
       locals: {
         recommendation: {
-          prisonOffender: { dateOfBirth: '2003-01-01' },
-          ppudOffender: { dateOfBirth: '1984-09-08' },
+          bookRecallToPpud: { dateOfBirth: '2003-01-01' },
         },
-        dateOfBirthNomis: '2003-01-01',
-        dateOfBirthPpud: '1984-09-08',
       },
     })
     const next = mockNext()
@@ -32,8 +29,6 @@ describe('get', () => {
       month: '01',
       year: '2003',
     })
-    expect(res.locals.dateOfBirthNomis).toEqual('2003-01-01')
-    expect(res.locals.dateOfBirthPpud).toEqual('1984-09-08')
     expect(res.render).toHaveBeenCalledWith('pages/recommendations/editDateOfBirth')
     expect(next).toHaveBeenCalled()
   })
--- a/server/controllers/recommendation/editDateOfBirthController.test.ts
+++ b/server/controllers/recommendation/editDateOfBirthController.test.ts
@@ -16,8 +16,11 @@ describe('get', () => {
     const res = mockRes({
       locals: {
         recommendation: {
-          bookRecallToPpud: { dateOfBirth: '2003-01-01' },
+          prisonOffender: { dateOfBirth: '2003-01-01' },
+          ppudOffender: { dateOfBirth: '1984-09-08' },
         },
+        dateOfBirthNomis: '2003-01-01',
+        dateOfBirthPpud: '1984-09-08',
       },
     })
     const next = mockNext()
@@ -29,6 +32,8 @@ describe('get', () => {
       month: '01',
       year: '2003',
     })
+    expect(res.locals.dateOfBirthNomis).toEqual('2003-01-01')
+    expect(res.locals.dateOfBirthPpud).toEqual('1984-09-08')
     expect(res.render).toHaveBeenCalledWith('pages/recommendations/editDateOfBirth')
     expect(next).toHaveBeenCalled()
   })
--- a/server/views/pages/recommendations/editDateOfBirth.njk
+++ b/server/views/pages/recommendations/editDateOfBirth.njk
@@ -25,6 +25,9 @@
                            {{ formatDateTimeFromIsoString({isoDate: recommendation.ppudOffender.dateOfBirth})  }}
                         {% endif %}
                     </div>
+                    <div id="date-of-birth-hint" class="govuk-hint">
+                        For example, 31 3 1980
+                    </div>
                     <div class="govuk-form-group{% if errors.dateOfBirth %} govuk-form-group--error{% endif %}">
                         {% if errors.dateOfBirth %}
                         <p id="error" class="govuk-error-message">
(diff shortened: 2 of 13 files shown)


server/controllers/recommendation/editDateOfBirthController.test.ts
```diff
@@ -16,8 +16,11 @@ describe('get', () => {
     const res = mockRes({
       locals: {
         recommendation: {
-          bookRecallToPpud: { dateOfBirth: '2003-01-01' },
+          prisonOffender: { dateOfBirth: '2003-01-01' },
+          ppudOffender: { dateOfBirth: '1984-09-08' },
         },
+        dateOfBirthNomis: '2003-01-01',
+        dateOfBirthPpud: '1984-09-08',
```
DarrenW-HMPPS, line 23: Are you sure you want to do this?  You set the value that you assert on later.  The controller could do nothing with `res` and the test would pass, I think.

server/views/pages/recommendations/editDateOfBirth.njk
```diff
@@ -11,23 +11,31 @@
         <div class="govuk-grid-column-two-thirds-from-desktop">
             {% include '../../partials/error-summary.njk' %}
 
-            {% set html %}
+            {# {% set html %}
```
DarrenW-HMPPS, line 14: Can we get rid of the commented code?

server/views/pages/recommendations/editDateOfBirth.njk
```diff
@@ -11,23 +11,31 @@
...
-                        For example, 31 3 1980
+                    <div class="govuk-body">
+                        {% if dateOfBirthNomis and dateOfBirthPpud %}
+                           <div class="govuk-summary-list__key">From NOMIS</div>
+                           {{ formatDateTimeFromIsoString({isoDate: dateOfBirthNomis})  }}
+                           <br><br>
+                           <div class="govuk-summary-list__key">From PPUD</div>
+                           {{ formatDateTimeFromIsoString({isoDate: dateOfBirthPpud})  }}
+                        {% else %}
+                            From NOMIS {{ dateOfBirthNomis }} <br>
```
DarrenW-HMPPS, line 37: It's best to avoid <br> if we can.  It's a blunt instrument that breaks the CSS styling approach.  It's caused us issues in the past.  I would think you can just delete it without any particular ill effects, but make sure things are wrapped in appropriately styled elements.

server/views/pages/recommendations/editDateOfBirth.njk
```diff
@@ -11,23 +11,31 @@
...
             <form novalidate method="post">
                 <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
-
                 <fieldset class="govuk-fieldset" role="group"
                           aria-describedby="date-of-birth-hint appointment-time-hint{% if errors.dateOfBirth %} error{% endif %}">
 
-                    <div id="date-of-birth-hint" class="govuk-hint">
-                        For example, 31 3 1980
+                    <div class="govuk-body">
+                        {% if dateOfBirthNomis and dateOfBirthPpud %}
```
DarrenW-HMPPS, line 30: Not convinced about this if...else logic.  It will display the "From NOMIS" line if dateOfBirthNomis is null.  Perhaps better to say `if dateOfBirthNomis - display dateOfBirthNomis `then separately,  `if dateOfBirthPppud - display dateOfBirthPpud`. That gives very clear and safe path through the code.  Or... if you want to make sure that "From NOMIS" is always displayed no matter what, don't include dateOfBirthNomis in any if check.

server/views/pages/recommendations/editDateOfBirth.njk
DarrenW-HMPPS, line 37: DateofBirthNomis needs formatting properly

DarrenW-HMPPS (changes requested):

server/controllers/recommendation/editDateOfBirthController.test.ts
bills7281: The recommendation is of course always present on `res.locals` in the recommendation route.
`res.locals.recommendation`.
>The controller could do nothing with res and the test would pass, I think.

I used the controller to grab a couple of fields from the recommendation and add two new `res.locals` fields
`dateOfBirthNomis` and `dateOfBirthPpud`. 
The test asserts that those two fields are present an has the correct value. If I removed `dateOfBirthNomis` and `dateOfBirthPpud` from `res.locals` in the controller the test will fail.

DarrenW-HMPPS: OK. So don't set the values in the test then.

server/views/pages/recommendations/editDateOfBirth.njk
bills7281: Done.

server/views/pages/recommendations/editDateOfBirth.njk
bills7281: Corrected as suggested 👍

server/views/pages/recommendations/editDateOfBirth.njk
bills7281: removed

bills7281: Corrected 👍

server/controllers/recommendation/editDateOfBirthController.test.ts
bills7281: The real `res.locals` receives the recommendation from the previous page in the rec flow. The `res.locals` in the test is a mock that needs to have it's values set.

bills7281: Done ... or rather undone! With a bit of a refactor! 👍

server/views/pages/recommendations/editDateOfBirth.njk
```diff
@@ -10,24 +10,20 @@
...
             <h1 class='govuk-heading-l'>{{ pageHeadings[page.id] }}</h1>
-
             <form novalidate method="post">
                 <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
-
                 <fieldset class="govuk-fieldset" role="group"
                           aria-describedby="date-of-birth-hint appointment-time-hint{% if errors.dateOfBirth %} error{% endif %}">
-
-                    <div id="date-of-birth-hint" class="govuk-hint">
-                        For example, 31 3 1980
```
DarrenW-HMPPS, line 30: This hint should be retained as it's part of the standard GDS date entry component.
It needs to be just above the input.
https://design-system.service.gov.uk/components/date-input

bills7281: image
Done!

DarrenW-HMPPS (approved):

xebu (approved):

PR #1067, opened 2024-01-24
mrd-2041_recall-received-date-validation-fix: date validation to enfo…

bills7281:
…rce past date only


server/controllers/recommendations/recallReceived/formValidator.test.ts
```diff
@@ -8,19 +8,19 @@ describe('validateRecallReceived', () => {
     basePath: `/recommendations/${recommendationId}/`,
     path: `/recommendations/${recommendationId}/check-booking-details`,
   }
-  const { year } = DateTime.now().plus({ years: 1 })
+  const { year } = DateTime.now().plus({ years: 0 })
```
DarrenW-HMPPS, line 11: Just `DateTime.now()` would do - no need to add 0

DarrenW-HMPPS (approved):

PR #1102, opened 2024-02-13
MRD-2307 - integrate release into booking on process

robinandrewdevilliers:


server/controllers/recommendation/bookToPpudController.ts
```diff
@@ -128,7 +140,7 @@ async function post(req: Request, res: Response, _: NextFunction) {
       dateOfSentence: nomisOffence.sentenceDate,
       licenceExpiryDate: nomisOffence.licenceExpiryDate,
       releaseDate: nomisOffence.releaseDate,
-      sentenceLength: { partDays: offenceTerm.days, partMonths: offenceTerm.months, partYears: offenceTerm.years },
+      sentenceLength: { partDays: offenceTerm?.days, partMonths: offenceTerm?.months, partYears: offenceTerm?.years },
```
DarrenW-HMPPS, line 143: This will fail when it gets to PPUD Automation if offenceTerm is null.  If offence term is null, sentenceLength can be null, but if sentenceLength is not null then it must have values.  The values can be 0.

server/controllers/recommendation/bookToPpudController.ts
```diff
@@ -137,6 +149,37 @@ async function post(req: Request, res: Response, _: NextFunction) {
...
+
+    await ppudUpdateRelease(token, offenderId, sentenceId, {
+      dateOfRelease: nomisOffence.releaseDate,
+      postRelease: {
+        assistantChiefOfficer: {
+          name: acoSigned.createdByUserFullName,
+          faxEmail: acoSigned.emailAddress,
+        },
+        offenderManager: {
+          name:
```
DarrenW-HMPPS, line 161: Nulls aren't allowed for name, faxEmail, telephone. Empty string is fine.

DarrenW-HMPPS (changes requested):

server/controllers/recommendation/bookToPpudController.ts
robindevilliers: not a problem.

server/controllers/recommendation/bookToPpudController.ts
robindevilliers: fixed

PR #1125, opened 2024-03-01
MRD-2177 - update ppud offender

robinandrewdevilliers:


server/controllers/recommendation/bookToPpudController.ts
```diff
@@ -83,53 +81,74 @@ async function post(req: Request, res: Response, _: NextFunction) {
...
+            croOtherNumber: recommendation.personOnProbation.croNumber,
+            dateOfBirth: recommendation.bookRecallToPpud?.dateOfBirth,
+            ethnicity: recommendation.bookRecallToPpud?.ethnicity,
+            familyName: recommendation.bookRecallToPpud?.lastName,
+            firstNames: recommendation.bookRecallToPpud?.firstNames,
+            gender: recommendation.bookRecallToPpud?.gender,
+            immigrationStatus: 'N/A',
+            nomsId: recommendation.personOnProbation?.nomsNumber,
+            prisonerCategory: 'N/A',
+            prisonNumber: recommendation.bookRecallToPpud?.prisonNumber,
```
DarrenW-HMPPS, line 142: We don't need to have any references to immigrationStatus or prisonerCategory in CaR, PPUD Automation deals with that as it is always the same value.

DarrenW-HMPPS: Same applies to youngOffender and status.

PR #1314, opened 2024-09-17
croNumber is a non-nullable string, and NOMIS' cro is nullable

MancRickster:


LucasCairns (approved):
LGTM 👍

MancRickster:
After investigating, the code shows some late changes to deprecate the use of a 'cro' field in favour of a 'croOtherNumber' field, although it's not entirely clear why.
Arguably changing the string-in-question's 'nullability' might be a better fix, but changing a data contract makes me a little nervous when I'm not totally familiar with the potential repercussions. 
```
data class SearchResultOffender(
  val id: String,
  @Deprecated("Replaced with croOtherNumber as CRO isn't specifically identifiable in PPUD")
  val croNumber: String,
  val croOtherNumber: String,
  ...
)
```

PR #1332, opened 2024-11-08
increasing ppud timeout

MancRickster:


Thomas-Geraghty:
Approving but short term solution, need proper solution 😄

Thomas-Geraghty (approved):

PR #1334, opened 2024-11-29
MRD-2619 Correct PPCS names capitalisation

ppmg:
We are sometimes getting names (first, middle & last) from other systems (e.g. NOMIS) in all-caps. With these changes we adjust the names in such cases before presenting them to the user.


MancRickster (approved):
lgtm

PR #1335, opened 2024-12-03
MRD-2600 Update recommendation status on SPO download

ppmg:
Currently, both POs and SPOs can create and download documents (Part A or DNTR), but only
if the former do it will the relevant recommendation's status be updated to indicate it
has been downloaded and sent to PPCS. This is an issue, as often (around 20% of the cases)
it's the SPO who is downloading and sending a Part A to PPCS, but the latter will only be
able to see a recommendation if it's status indicates its document has been downloaded.

The reason behind the current behaviour is that the official process dictates that, once
they have countersigned a recommendation, the SPO should alert the PO, who is then
responsible for creating and downloading the document. There are plans to review the
screens to perhaps, in the long term, enforce said process. Until then, the changes
introduced with this commit will continue to allow SPOs to create and download them, and
will also mark the recommendations as having their document downloaded and setn to PPCS,
so that PPCS users can continue on their part of the journey.


MancRickster (approved):
lgtm

PR #1336, opened 2024-12-05
Upgrade cross-spawn

ppmg:
Our project pulls in cross-spawn as an indirect dependency. The version we are pulling in has a known CVE vulnerability, so we're ugprading it to the latest known version without it.


MancRickster (approved):
lgtm

PR #1340, opened 2024-12-19
Mrd 2529 display new osp values

ppmg:
ARN switched to a different set of OSP values, OSP/DC and OSP/IIC.
With this change, the new values are displayed instead of the old
ones if present, with the relevant labels updated accordingly. If
neither new nor old values are present, the new labels are
displayed above the 'Data not available.' messages.

The timeline will also reflect the new and old values.


MancRickster (approved):
LGTM

PR #1341, opened 2024-12-23
MRD-2529 Fix OSP/C & OSP/DC label displays

ppmg:
The check on the level value was incorrect for OSP/C and OSP/DC values in the
case of a value of VERY_HIGH: the value itself is 'VERY_HIGH', with an
underscore, while the value checked against was 'VERY HIGH', with a space in
between. This was resulting in the label marker for VERY_HIGH cases to be
displayed in the middle of the rating bar, rather than on the right-most
section, due to the label not having the right class assigned to it. This is
corrected here.

It was also noticed that the wrong CSS class was assigned to the display label
for LOW OSP/C and OSP/DC scores: the one corresponding to the MEDIUM one
was used instead. This is corrected here.


MancRickster (approved):
lgtm

PR #1343, opened 2025-01-03
correcting a test's hardcoded reference to the year 2024

MancRickster:
recallReceieved formValidator expected a hardcoded resultant year of 2024 despite the test's input values using the current year. Year also reduced by 1 since the validation expects the date to be in the past.


ppmg (approved):
LGTM

PR #1346, opened 2025-01-07
Bump cypress multi reporters nanoid vuln CVE 2024 55565 force mocha bump

MancRickster:
There isn't a version of cypress-multi-report that depends on a version of mocha which doesn't have a nanoid dependency, so this forces c-m-r to use the next version of mocha which drops that dependency.
More info on c-m-r / mocha reporters:
https://mochajs.org/#reporters

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10561,9 +10561,9 @@
       }
     },
     "node_modules/mocha": {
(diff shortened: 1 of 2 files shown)


ppmg (approved):
LGTM

PR #1347, opened 2025-01-07
mrd 2612 display graceful message when mappa value unknown

MancRickster:
Given a PPCS user is Booking On a PoP using CaR
When they visit the ‘Edit MAPPA’ page and the MAPPA does not have a value
Then display the following:

From the Part A

MAPPA level: Unknown


ppmg (approved):
LGTM

PR #1348, opened 2025-01-08
MRD-2609 Set 'Sentenced Under' field in PPUD Sentence

ppmg:
The existing 'Released Under' field for PPUD Offender Releases is now re-used
also for PPUD Offender Sentences' 'Sentenced Under' field. This is reflected
both in its storage as a separate, new legislationSentencedUnder field in the
bookRecallToPpud structure as well as in the requests for creating and updating
PPUD sentences, as the legislation under which an offender was sentenced should
be the same as the one under which they were released.


MancRickster (approved):
LGTM

PR #1349, opened 2025-01-20
mrd-2653 Cache PPUD user hits

MancRickster:
Adding caching of PPUD user api response for 1 week
Including additional override parameter for items that should remain cached for longer than the default (1d)

### DIFF
--- a/server/controllers/startPage/startPage.test.ts
+++ b/server/controllers/startPage/startPage.test.ts
@@ -1,3 +1,4 @@
+import { createClient } from 'redis'
 import { mockReq, mockRes } from '../../middleware/testutils/mockRequestUtils'
 import { startPage } from './startPage'
 import { ppudSearchActiveUsers, searchMappedUsers } from '../../data/makeDecisionApiClient'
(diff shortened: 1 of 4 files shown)


server/data/fetchFromCacheOrApi.ts
```diff
@@ -9,6 +9,7 @@ interface Fn {
     checkWhetherToCacheDataFn: (data: Type) => boolean
     userId: string
     redisKey: string
+    ttlOverrideSeconds?: number
```
ppmg, line 12: Could you add a test to check this new parameter is working as expected, please?

server/controllers/startPage/startPage.test.ts
```diff
@@ -1,12 +1,30 @@
...
 import { startPage } from './startPage'
 import { ppudSearchActiveUsers, searchMappedUsers } from '../../data/makeDecisionApiClient'
 import searchMappedUsersApiResponse from '../../../api/responses/searchMappedUsers.json'
 import ppudSearchActiveUsersApiResponse from '../../../api/responses/ppudSearchActiveUsers.json'
 
 jest.mock('../../data/makeDecisionApiClient')
+jest.mock('redis')
 
 describe('startPage', () => {
+  const redisGet = jest.fn()
```
ppmg, line 12: The `fetchFromCacheOrApi` component is tested on its own, could we not mock it here instead of the `redis` stuff it does underneath? Otherwise if we ever change the internals of the component we'll have to change this test too.

MancRickster: Updated in next commit.

server/data/fetchFromCacheOrApi.ts
MancRickster: Added in next commit.

ppmg (approved):
LGTM

PR #1351, opened 2025-01-27
MRD-2611 Fix autocomplete issues

ppmg:
Three issues with the autocomplete component have been addressed:
 - Issue with behaviour on item selection with invalid input.
 - Placeholder text overlap.
 - Menu item font.

Item Selection
The autocomplete field wasn't behaving as we want it to: if a valid option was
set at any point, the component would remember it even if an invalid value was
provided afterwards (e.g. select a valid option, then type in an alternative
search which doesn't produce any results). This resulted in the remembered
value being set if the relevant form was submitted, even if it wasn't being
displayed to the user at the time.

This has been fixed by checking every time the field loses focus: if what is
currently typed into the field doesn't correspond to any of the available
options, the empty value option will be selected in the underlying select
element.

Placeholder text
The autocomplete placeholder doesn't work well in the autocomplete component if
combined with the 'autoselect' option. With the latter configured, as soon as
you click on the input field it will display in a faint font which option will
be set if the 'Enter' key is pressed (if you set a minimum query length I
expect this issue doesn't occurs, but we aren't setting a minimum length).
Since the placeholder doesn't vanish until you start typing, the placeholder
text and that coming from the autoselect functionality overlap in the same
field.

This is addressed here via CSS: whenever the placeholder is focused on, it is
made invisible.

Menu item font
For unknown reasons, the menu items of the autocomplete component didn't have
the expected, standard gov uk font family applied. A list item class has been
explicitly set on the menu items (via an autocomplete parameter) to address this
(but can be removed if the cause of the font not being inherited is ever identified
and addressed).


MancRickster (dismissed):
LGTM

MancRickster (approved):
LGTM

PR #1352, opened 2025-01-28
MRD-2526 Fixing info endpoint (move to standard port)

MancRickster:
Reverting management endpoints (health/info/prometheus) back to standard HTTP ports.
These were overridden to use 8081 on 03/05/22 by Darren Oakley for MRD-11


ppmg (approved):
LGTM

PR #1353, opened 2025-01-28
MRD-2611 Update autocomplete placeholder text

ppmg:
The placeholder text in the two autocomplete fields has been changed from
"Select X" to "Search for X", as the latter has a stronger implication or
suggestion of the field allowing the user to input text to search for the
option they want to select, rather than just having to scroll down to find the
appropriate option. The words "Type" and "Write" were also considered, but
discarded due to not being applicable in all cases (e.g. when certain
accessibility input devices are being used by the user).


MancRickster (approved):
LGTM

PR #1355, opened 2025-02-07
MRD-2608 Adding optional text for ppud index offence

MancRickster:
Allow optional text comment when entering ppud index offence

### DIFF
--- a/server/controllers/recommendation/matchIndexOffenceController.test.ts
+++ b/server/controllers/recommendation/matchIndexOffenceController.test.ts
@@ -74,6 +74,7 @@ describe('post', () => {
       params: { recommendationId: '1' },
       body: {
         indexOffence: 'some offence',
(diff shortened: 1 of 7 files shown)


server/controllers/recommendation/matchIndexOffenceController.test.ts
```diff
@@ -144,6 +146,7 @@ describe('post', () => {
         bookRecallToPpud: {
           policeForce: 'Kent',
           indexOffence: 'some offence',
+          indexOffenceComment: null,
```
ppmg, line 149: Can you add tests for the empty string and whitespace cases too, please, to make sure it's setting `indexOffenceComment` to `null` when expected?

ppmg (changes requested):

ppmg (approved):
LGTM

PR #1357, opened 2025-02-07
MRD-2610 Minutes subject fix

MancRickster:
Ppud aut facilitates creation of minutes without Subjects, but blocks incoming calls with the same. Subject now supplied


ppmg (approved):
LGTM

PR #1358, opened 2025-02-18
MRD-2682 police contact details reset if

MancRickster:
The Police Contact Details are reset/nulled if the Custody Status is changed from “Yes, police custody” to either of the other two options.

Original Aug '22 logic: MRD-559: DEV: If PoP in custody, don't show arrest issues / police contact pages
Updated Feb '24 logic: MRD-2391: [Approved] Stop selecting 'in custody' hiding police contact info from task list

According to the second card, the contact details should not be reset/nulled.


ppmg (approved):
LGTM

PR #1359, opened 2025-02-19
Mrd 2639 Add current establishment to PPCS journey

ppmg:
When a PPCS user books an offender into PPUD, they fill in the offender's
current establishment. At the moment, this happens manually, but it is going to
be incorporated into the CaR journey, so we need to make these changes here so
the establishment can be stored during said journey and passed through to PPUD
at booking time.

### DIFF
--- a/server/controllers/recommendation/ppcs/currentEstablishment/recommendationEstablishmentExtractor.test.ts
+++ b/server/controllers/recommendation/ppcs/currentEstablishment/recommendationEstablishmentExtractor.test.ts
@@ -19,15 +19,15 @@ describe('extractNomisEstablishment', () => {
   })
 
   it('extracts non-null establishment', () => {
-    recommendation.prisonOffender.agencyId = 'abc'
+    recommendation.prisonOffender.agencyDescription = 'abc'
 
     const actualNomisEstablishment = extractNomisEstablishment(recommendation)
 
-    expect(actualNomisEstablishment).toEqual(recommendation.prisonOffender.agencyId)
+    expect(actualNomisEstablishment).toEqual(recommendation.prisonOffender.agencyDescription)
   })
 
   it('extracts undefined establishment', () => {
-    delete recommendation.prisonOffender.agencyId
+    delete recommendation.prisonOffender.agencyDescription
 
     const actualNomisEstablishment = extractNomisEstablishment(recommendation)
 
--- a/server/controllers/recommendation/ppcs/currentEstablishment/recommendationEstablishmentExtractor.ts
+++ b/server/controllers/recommendation/ppcs/currentEstablishment/recommendationEstablishmentExtractor.ts
@@ -1,7 +1,7 @@
 import { RecommendationResponse } from '../../../../@types/make-recall-decision-api'
 
 export const extractNomisEstablishment = (recommendation: RecommendationResponse): string => {
-  return recommendation.prisonOffender?.agencyId
+  return recommendation.prisonOffender?.agencyDescription
 }
 
 export const extractPpudEstablishment = (recommendation: RecommendationResponse): string => {


integration_tests/integration/recommendation.spec.ts
```diff
@@ -2266,6 +2266,29 @@ context('Make a recommendation', () => {
...
+        response: {
+          ...completeRecommendationResponse,
+          prisonOffender: {},
+          bookRecallToPpud: {
+            currentEstablishment: 'The Kyln',
+          },
+          ppudOffender: {},
+        },
+      })
+      cy.task('getStatuses', { statusCode: 200, response: [{ name: 'SENT_TO_PPCS', active: true }] })
```
KiraNamida, line 2281: I have seen this `'SENT_TO_PCS'` status appearing in a bunch of places. Should this be a const somewhere?

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
```diff
@@ -22,6 +26,8 @@ const PRISON_OFFENDER_TEMPLATE = {
   lastName: prisonOffenderLastName,
   facialImageId: 1234,
   dateOfBirth: '1970-03-15',
+  agencyId: 'KLN',
+  agencyDescription: 'The Kyln',
```
KiraNamida, line 30: This is another field I've seen in a few places. Should we have this as codified test data somewhere?

server/views/pages/recommendations/editCurrentEstablishment.njk
```diff
@@ -0,0 +1,58 @@
...
+
+            <h1 class='govuk-heading-l'>{{ pageHeadings[page.id] }}</h1>
+
+            <div class='govuk-body'>
+                <div class="govuk-text-list">
+                    <h2 class='govuk-heading-m'>From NOMIS</h2>
+                    <div class="govuk-text-list__row">
+                        <div class="govuk-text-list__key">Establishment</div>
+                        <div class="govuk-text-list__value"
+                             data-qa="nomisEstablishment">{{ nomisEstablishment }}</div>
```
KiraNamida, line 22: I've been seeing a lot of these data-qa tags about the place in use for retrieving elements in cypress tests. Is there a current argument for why this is being done, rather than say meaningful ids which could double up as fragment links?
Not that it's critical but I have always considered this a last resort as it adds additional markup in places like production where it not necessarily relevant.

integration_tests/integration/recommendation.spec.ts
ppmg: Good point! It is already a const somewhere, I'll update this and the rest of the similar cases in this source file.

server/views/pages/recommendations/editCurrentEstablishment.njk
KiraNamida: It's more a good practice argument from certain circles than a functional one.
The primary debate is that it adds elements that are not strictly required nor that describe what the element actually is. From reading this in a test file alone the elemtn with `[data-qa="nomisEstablishment"]` could be anything: an input? a label? a table cell? a definition?
The alternative is to write selectors that step through the element and class/id. Presuming the above was given `id=nomisEnhancement` you could write the selector (and I would have to test this as writting on the cuff here =P) like `get('div[id^="nomisEnhancement"][class^="govuk-text-list-value"])` and you know you are dealing with a list value.
This is not the greatest example as it is plain text output so probably has the least impact in this case but this can start to make waves when you begin getting things like inputs, or you start getting by defined elements like headings.

KiraNamida: This is not to say that data attributes are never useful, they where originally conceived to pick out dynamic elements in a page.
The classic example is a counter on a page that the user can change. This is a small isolated element that might want to be interrogated regularly so you might choose to wrap the actual value in a span and tag it with a data attribute that can de directly retrieved (`data-id` or `data-test-id` are common use cases).
This breaks the above because there is a specific need to pick out a part of the copy of the page, rather than a domain element that have preexisting ways to identify them.
As I said, this is not make or break. Users are not likely to start digging through the html and complain we have data tags for testing in the domain but it got my gears turning when I started seeing this and having dug into git more I can see it was standard practice for testing from way back when,

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
ppmg: No, these come from a system external to our team - there is an endpoint to get all the values we are interested in, but the API for the endpoint doesn't pre-define what the full set of possible values is.

server/views/pages/recommendations/editCurrentEstablishment.njk
ppmg: That all makes sense, thanks for the explanation! As you said, we can look into changing this separately once there's time for it, so as not to expand this PR's scope more than is needed.

integration_tests/integration/recommendation.spec.ts
ppmg: The project wouldn't build for some reason I couldn't pinpoint when importing said enum from `recommendationStatusCheck.ts`. Since the enum isn't used within that source, I've made a copy of it into a separate source and imported that, which worked (which suggests an issue with what else it pulls in form `recommendationStatusCheck.ts` when doing the import, but I couldn't find out the cause). I've left the old enum around too, as it's referenced by too many files to move over with this PR, but have opened https://dsdmoj.atlassian.net/browse/MRD-2688 for this.

integration_tests/integration/recommendation.spec.ts
```diff
@@ -2266,6 +2266,29 @@ context('Make a recommendation', () => {
...
+      cy.task('getReferenceList', {
+        name: 'establishments',
+        statusCode: 200,
+        response: { values: ['The Kyln'] },
+      })
+
+      cy.visit(`/recommendations/252523937/edit-current-establishment`)
+      cy.pageHeading().should('contain', 'Edit current establishment')
```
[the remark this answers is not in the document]
ppmg: I think that might be what other earlier tests in this same source file intended to be, but the intention might have been lost over time as tests were copy-pasted and adjusted.

I assumed there was a bit of navigation that would be going on for all PPCS cases before each individual test is run, but that was a wrong assumption. We don't need most of the prep that was here, so I've done away and left only what is strictly needed, which is the mocking of the call to get the reference list of establishments, as that is called by the controller for this page.

I think we might want to discuss at some point what the scope of these tests should be: should they just be opening the relevant page and check they load with the minimum required data? Should they do a bit more (e.g. for this current establishment test, `res.locals` normally would have a recommendation and a couple of fields would be filled in with data from it, but if there's no recommendation (as is the case here, since the test case doesn't navigate through the page which puts the recommendation into `res.locals`), the page still loads fine)?

KiraNamida (approved):
LGTM :shipit:

MancRickster (approved):
LGTM

PR #1360, opened 2025-02-28
MRD-2639 Add establishment to PPUD requests

ppmg:


server/booking/bookOffender.test.ts
```diff
@@ -272,6 +280,7 @@ describe('book offender', () => {
         firstNames: 'Johnny J',
         lastName: 'Teale',
         dateOfBirth: '1970-03-15',
+        currentEstablishment: establishment,
```
MancRickster, line 283: Is there a reason why currentEstablishment is used here instead of continuing with just 'establishment' (or vica versa)?

ppmg: I've tried to keep each field in line with what the data owner of the context it's a part of calls it - so `agency` for `prisonOffender` and other data structures relating to or coming from NOMIS, `establishment` for `ppudOffender` and other PPUD ones and `currentEstablishment` for `bookRecallToPpud` and other CaR ones.

MancRickster (approved):
LGTM

PR #1361, opened 2025-03-03
MRD-2536 Maintenance page and Notifications banner fixes

MancRickster:
The current maintenance page redirect code in app.ts is above the middleware handlers for static resources, so the stylesheets (and other stuff) isn’t being served correctly. 

    The current maintenance page code is either on/off depending on two  env. vars. IMO it would be better if this feature was also date-controlled; so we should probably setup new Start and End date parameters for this functionality


server/utils/utils.test.ts
```diff
@@ -232,44 +231,59 @@ describe('validateCrn', () => {
...
-    config.notification.endDate = null
-    const result = isBannerDisplayDateRangeValid()
+    const result = isDateTimeRangeCurrent('2023-09-18', null)
     expect(result).toEqual(false)
   })
+  it('returns false when end date is an invalid date', () => {
+    const result = isDateTimeRangeCurrent('2050-09-13', 'invalid string')
+    expect(result).toEqual(false)
+  })
+  it('returns true when the current date and time is within the specified date and time range', () => {
```
ppmg, line 267: Could we have another test like this to check out of range, please?

server/views/pages/startPage.njk
```diff
@@ -2,24 +2,25 @@
...
+        {{ govukNotificationBanner({html: html}) }}
+    {% endif %}
 {% endblock %}
 {% block content %}
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds-from-desktop">
             <h1 class="govuk-heading-xl">{{ applicationName }}</h1>
 
-            <p>Use this service to view information about a person on probation. This will help you make a recommendation about recall.</p>
+            <p>Use this service to view information about a person on probation. This will help you make a
```
ppmg, line 22: Is this a deliberate line break or an auto-format one? I think we had an issue recently where we'd accidentally broken a line that we shouldn't have, although I can't recall where that was or if it was only an issue in that case.

ppmg (changes requested):

server/views/pages/startPage.njk
KiraNamida: Unsure what the previous problem was but seeing as this is plain text within a <p> element, the text will be rendered linearly by the browser so line breaks would not be an issue, unless there is something else specific to this application that is creating an exception to this.

server/views/maintenance-page.njk
```diff
@@ -2,25 +2,30 @@
...
 {% endblock %}
 
 {% block content %}
-{{ super() }}
+    {{ super() }}
 
 
-<div class="govuk-grid-row">
+    <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
```
KiraNamida, line 18: Should this be intended again as well?

server/app.ts
```diff
@@ -40,24 +42,37 @@ export default function createApp(userService: UserService): express.Application
...
   app.use(appInsightsOperationId)
-
   app.use(metricsMiddleware)
   app.use(setUpHealthChecks())
   app.use(setUpWebSecurity())
   app.use(setUpWebSession())
   app.use(setUpWebRequestParsing())
   app.use(setUpStaticResources())
+
+  if (isDateTimeRangeCurrent(config.maintenancePage.startDateTime, config.maintenancePage.endDateTime)) {
```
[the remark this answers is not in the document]
ppmg: I'm not sure, I haven't seen a maintenance period since I joined, will need to hear from @MancRickster for that.

I'm not sure I understand what you mean by the app pools.

server/views/pages/startPage.njk
MancRickster: Yer, auto-format chaff

ppmg:
LGTM, but would appreciate another review from @KiraNamida in case he sees any issues with the new approach.

KiraNamida (approved):
Looking good. :shipit: from me.

PR #1367, opened 2025-04-08
MRD-2700 Handle movement retrieval 404 error

ppmg:
Logic is added here to handle 404 errors when retrieving offender movements
during the establishment mapping calculations. Such an error is unlikely to
happen, as we only retrieve movements if the offender is marked as being
mid-transfer, which is meant to guarantee there is a corresponding movement.
However, if the Prison API system is down or corrupted, we need to handle
this in a way that doesn't block the user.

Given the unlikelihood of such a use case, and the uncertainty of how we should
react without knowing why we got a 404 (can we make further calls to Prison API
or is it now unreliable?), we leave the field empty for the user to make a more
informed decision.

Additionally, the check-booking-details test is here updated to include an
agencyId as part of the offender's details, which in turn triggers the retrieval of
the establishment mappings in order to determine the relevant PPUD
establishment. Although not strictly necessary for the page to load, the most
common case will be for there to be an agencyId and for it to be mapped, so we
exercise that code here.


KiraNamida (approved):
:shipit:

PR #1375, opened 2025-05-20
MRD-2744 Remove pre-prod e2e tests

ppmg:
The pre-prod e2e test run needs to be removed, as the data
originates in prod and shouldn't be used for anything other
than its original purpose (GDPR).

Additionally, MFA has just been added to pre-prod, which
would prevent the automated tests from going through anyway.


KiraNamida (approved):
:shipit:

PR #1376, opened 2025-05-20
MRD-2743 Update dependencies

ppmg:
The following dependencies have been updated to address vulnerabilities:
  - babel to ^7.26.10 - the version has been overridden, as the consumer of it,
  ts-jest, doesn’t pull in a high enough version even in its latest release.
  No breaking changes are listed in the babel change for our update range.
  - mocha to ^10.6.0 - also overridden due to the consumers
  (cypress-multi-reporters and mocha-unit-reporter) not up-to-date even on
  their latest versions. No breaking changes are listed for our update range.
  - super agent to ^10.0.0 - this was to bring up the version of formidable,
  one of its dependencies which had a vulnerability. No breaking changes have
  been listed since super agent released 6.1.

Note also that the ts-jest dependency version expression has been changed from
^29.1.2 to ~29.1.2. This is because that project explicitly states they don’t
follow server. From the look of it, they use minor versions for breaking
changes, but I’m not sure if that always holds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "prom-client": "^15.1.1",
     "redis": "^4.6.13",
     "striptags": "3.2.0",
-    "superagent": "^9.0.0",
+    "superagent": "^10.0.0",
     "ts-loader": "^9.5.1",
     "url-value-parser": "^2.2.0",
     "uuid": "^10.0.0",
@@ -202,7 +202,7 @@
     "prettier": "^3.2.5",
     "sass": "^1.74.1",
     "supertest": "^7.0.0",
-    "ts-jest": "^29.1.2",
+    "ts-jest": "~29.1.2",
     "typescript": "^5.6.3",
     "wait-for-expect": "^3.0.2",
     "xml2js": "^0.6.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -202,7 +202,7 @@
     "prettier": "^3.2.5",
     "sass": "^1.74.1",
     "supertest": "^7.0.0",
-    "ts-jest": "^29.1.2",
+    "ts-jest": "~29.1.2",
```
KiraNamida, line 205: Are we confident about future minor patches on this?

KiraNamida (commented):
:shipit:

package.json
ppmg: As discussed, it seems I might have misinterpreted the statement in the Versioning section in https://www.npmjs.com/package/ts-jest about them not following semver. What they mean by that might be that they might release minor versions without new features and major ones without breaking changes in order to stay aligned version-wise with jest. However, since the change here has been to a more restrictive pattern, and we should be checking for breaking changes whenever libraries are updated anyway, we've agreed to go ahead with this as is.

PR #1377, opened 2025-05-20
Update fix: paging for search results - by crn vs name

KiraNamida:
To reflect the changes in https://github.com/ministryofjustice/make-recall-decision-api/pull/1179

As the split in search endpoints has lead to different assumptions as the the index of the first page (by crn: 1, by name: 0) the managing of this has been moved into the API as an implementation details and now the UI and API will always communicate on the understanding that the first page has index 1.


LucasCairns (approved):
LGTM 👍

PR #1390, opened 2025-07-23
Mrd 2734 edit release date details

EmilyHazlehurst:


integration_tests/integration/recommendations/ppcs/indeterminateSentences/sentenceToCommitIndeterminate.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+        s => s.id === defaultRecommendationResponse.bookRecallToPpud.ppudSentenceId
+      )
+      beforeEach(() => {
+        cy.task('getRecommendation', { statusCode: 200, response: defaultRecommendationResponse })
+        cy.task('getStatuses', { statusCode: 200, response: defaultPPCSStatusResponse })
+      })
+      it('displays the default recommendation data', () => {
+        const summaryListContent = {
+          rows: [
+            { key: 'Custody type', value: defaultRecommendationResponse.bookRecallToPpud.custodyGroup },
```
ppmg, line 49: This should be selecting the custodyType, not the custodyGroup.

server/helpers/ppudSentence/custodyTypes.ts
```diff
@@ -1,6 +1,7 @@
 export const determinateCustodyTypes = ['Determinate', 'EDS', 'EDS (non parole)'] as const
 
 export const indeterminateCustodyTypes = [
+  'Indeterminate',
```
ppmg, line 4: This needs to be removed, as it isn't an existing type (just a group).

ppmg (changes requested):

integration_tests/integration/recommendations/ppcs/indeterminateSentences/sentenceToCommitIndeterminate.spec.ts
EmilyHazlehurst: @ppmg if I change it to custodyType it returns `undefined` 

In ppudSentenceGenerator.ts there's a PpudSentenceOptions type that has custodyType on it, but that doesn't include Indeterminate which is why I added it initially.

There's this enum that suggests the group is determinate/indeterminate:

`export enum CUSTODY_GROUP {
  DETERMINATE = 'Determinate',
  INDETERMINATE = 'Indeterminate',
}
`

So I'm wondering if actually it _should_ be custodyGroup, but that `PpudSentenceOptions` should have the custodyGroup instead of type? Or have we just been inconsistent with our naming?

Because the determinateCustodyTypes are _determinate_, _eds_ and _eds (non-parole)_ but I wouldn't expect _determinate_ to appear in that list if it's the umbrella group?

EmilyHazlehurst: Ah, I think we're displaying the wrong information in the controller. We're using CustodyGroup but looking at the designs it should be CustodyType as the example is a type, not a group.

PR #1393, opened 2025-08-01
MRD-2775 FTR48 mandated FTR path

ppmg:
The recall-type page is updated to support the path where an FTR
is mandated due to the answers given to the suitability questions.


KiraNamida (approved):
A small high level improvement not related to the immediate functionality of this ticket and some type stuff to look at.
In @ppmg's absence I would happy to merge and test this and then look at them as my next work needs to extend what is here as it is.

PR #1394, opened 2025-08-04
Mrd 2774 update car to reflect ftr suitability criteria

EmilyHazlehurst:


KiraNamida (commented):
For when the FTR48 feature flag is enabled, have we confirmed what will happen with recommendations made using the old version of the screen should the CA pull them up using the new screen?
Will values essentially being dropped and the validation of this page enforcing the new set of values be acceptable from the client team?

KiraNamida:
> For when the FTR48 feature flag is enabled, have we confirmed what will happen with recommendations made using the old version of the screen should the CA pull them up using the new screen? Will values essentially being dropped and the validation of this page enforcing the new set of values be acceptable from the client team?

So, I started reviewing [this PR](https://github.com/ministryofjustice/make-recall-decision-ui/pull/1393) following this one which handles the mandated FTR on the following page and I see this does introduce this so that answers that one.

EmilyHazlehurst:
> For when the FTR48 feature flag is enabled, have we confirmed what will happen with recommendations made using the old version of the screen should the CA pull them up using the new screen? Will values essentially being dropped and the validation of this page enforcing the new set of values be acceptable from the client team?

@KiraNamida I'm not sure on that one yet as I'm not 100% what the pilot/demo release is going to look like. I'll find out today. 

Longer term there's MRD-2804 which covers deletion of all _in progress_ recommendations on release.

PR #1395, opened 2025-08-06
Add non mandatory (discretionary) panel to recall type for FTR48

KiraNamida:
[MRD-2786 Change recommendation screen - suitability for FTR [Discretionary path] or Standard](https://dsdmoj.atlassian.net/browse/MRD-2786)

Add discretionary panel to the recall type controller if the FTR48 feature is enabled and any of the exclusion criteria (values from the suitability for recall type page) apply.
Else continue to display the mandatory panel.

image

### DIFF
--- a/integration_tests/integration/recommendations/partA/recallType.spec.ts
+++ b/integration_tests/integration/recommendations/partA/recallType.spec.ts
@@ -1,0 +1,249 @@
+import { fakerEN_GB as faker } from '@faker-js/faker'
+import { testForErrorPageTitle, testForErrorSummary } from '../../../componentTests/errors.tests'
+import { RecommendationResponseGenerator } from '../../../../data/recommendations/recommendationGenerator'
+
+context('Recal Type Page', () => {
+  const testPageUrl = `/recommendations/123456789/recall-type`
+
+  const defaultRecommendation = RecommendationResponseGenerator.generate({
+    recallType: 'none',
+  })
+
+  beforeEach(() => {
+    cy.task('getRecommendation', { statusCode: 200, response: defaultRecommendation })
+    cy.task('getStatuses', { statusCode: 200, response: [] })
+    cy.signIn()
+  })
+
+  describe('Page Data', () => {
+    it('Standard page load', () => {
+      cy.visit(testPageUrl)
+
+      // Page Heading
+      cy.pageHeading().should('equal', 'What do you recommend?')
+
+      // Recall type Radio inputs
+      cy.get('.govuk-form-group').should('have.length', 3).eq(0).as('recallTypeFormGroup')
+
+      cy.get('@recallTypeFormGroup').get('fieldset').should('exist').as('recallTypeFieldset')
+
+      cy.get('@recallTypeFieldset').get('legend').should('exist').should('contain.text', 'Select your recommendation')
+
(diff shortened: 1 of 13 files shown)


integration_tests/integration/recommendations/partA/recallType.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+          isMappaLevel2Or3: randomCriteria.isMappaLevel2Or3,
+          isRecalledOnNewChargedOffence: randomCriteria.isRecalledOnNewChargedOffence,
+          isServingFTSentenceForTerroristOffence: randomCriteria.isServingFTSentenceForTerroristOffence,
+          hasBeenChargedWithTerroristOrStateThreatOffence:
+            randomCriteria.hasBeenChargedWithTerroristOrStateThreatOffence,
+        })
+        const expectedName = recommendationWithExceptionCriteria.personOnProbation.name
+
+        describe(`Randomised Criteria: 48+:${randomCriteria.isSentence48MonthsOrOver} | 18+:${randomCriteria.isUnder18} | M4:${randomCriteria.isMappaCategory4} | M2/3:${randomCriteria.isMappaLevel2Or3} | New:${randomCriteria.isRecalledOnNewChargedOffence} | STerr:${randomCriteria.isServingFTSentenceForTerroristOffence} | CTerr:${randomCriteria.hasBeenChargedWithTerroristOrStateThreatOffence}`, () => {
+          it('The the discretionary suitability panel is displayed', () => {
```
EmilyHazlehurst, line 205: Typo, repeated 'the'

server/testUtils/booleanUtils.ts
```diff
@@ -0,0 +1,15 @@
+export const generateBooleanCombinations = (n: number) => {
```
EmilyHazlehurst, line 1: Could you add a function definition to this? It doesn't need types since we're using TypeScript, but it would be good to have a description and usage example e.g. 
```
/**
 * Generates boolean combinations of a specified length
 *
 * @param n - The length of each boolean combination to generate
 * @returns An array of boolean arrays, each intended to be length `n`
 * 
 * @example
 * generateBooleanCombinations(2)
 * // Might return:
 * // [
 * //   [true, true],
 * //   [true, false],
 * //   [false, true]
 * // ] 

```

EmilyHazlehurst (changes requested):
@KiraNamida a few small things, but otherwise looks good 😄

integration_tests/integration/recommendations/partA/recallType.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+          isUnder18: false,
+          isMappaCategory4: false,
+          isMappaLevel2Or3: false,
+          isRecalledOnNewChargedOffence: false,
+          isServingFTSentenceForTerroristOffence: false,
+          hasBeenChargedWithTerroristOrStateThreatOffence: false,
+        })
+        const expectedName = recommendationWithoutExceptionCriteria.personOnProbation.name
+
+        it('The the mandatory suitability panel is displayed', () => {
```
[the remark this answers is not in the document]
KiraNamida: Ah, that should the `Then the`, but a typo all the same. 🤦

EmilyHazlehurst (approved):
@KiraNamida I've left a comment on the randomiseCriteria() function where it can be simplified a bit, but otherwise all looks good 👍

PR #1398, opened 2025-08-15
Mrd 2788 display a banner for ftr48 changes

EmilyHazlehurst:


KiraNamida (approved):
Some nice simplifications in boiler plater code. 👌
Are there any plans for config driven setting for the notifications? Looking at how this will require a full pipeline deployment to update which should be fine for this deadline but wondering about clean up or sudden notices in the future.
Though I am guessing the answer is "currently out of scope"? 🤔

EmilyHazlehurst:
> Some nice simplifications in boiler plater code. 👌 Are there any plans for config driven setting for the notifications? Looking at how this will require a full pipeline deployment to update which should be fine for this deadline but wondering about clean up or sudden notices in the future. Though I am guessing the answer is "currently out of scope"? 🤔

Thanks @Dan

So there’s still the existing banner in there which can be turned on/off with the environment variables, this one just had added the added complexity of different banners being needed for different user roles. 

I’ve made a ticket to review the banners as there are too many on various pages, so although out of scope for this one we can explore that as part of that work.

PR #1399, opened 2025-08-15
FTR48 Fixups: Recall Type error message

KiraNamida:
Modulate noRecallType error messages into Mandatory and Discretionary variants

Discretionary recall error message referring to all three options: Standard, Fixed Term and no recall
Discretionary Error Message

Update Mandatory recall error message referring only to Fixed Term and No Recall
Mandatory Error Message

Additionally, change the order of `Fixed Term` and `Standard` Recall options as UCD wishes Fixed Term to be the presumed default and offered as the lead option.
This matches the designs in [MRD-2775](https://dsdmoj.atlassian.net/browse/MRD-2775) but it was not apparent that the order of the inputs mattered until we noticed that the error message listed them in a different order so they were simply in the order they have always been.

[MRD-2775]: https://dsdmoj.atlassian.net/browse/MRD-2775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration_tests/integration/recommendationFormValidation.spec.ts
+++ b/integration_tests/integration/recommendationFormValidation.spec.ts
@@ -154,7 +154,7 @@ context('Make a recommendation - form validation', () => {
     cy.clickButton('Continue')
     cy.assertErrorMessage({
       fieldName: 'recallType',
-      errorText: "Select if you're recommending a fixed term recall, standard recall, or no recall",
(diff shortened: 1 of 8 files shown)


ppmg (approved):
LGTM

PR #1400, opened 2025-08-15
FTR48 Fixups: 

KiraNamida:
Only include Recall type rational when rendering for pre-FTR48 or when the decision is discretionary.

None FTR48 Version: maintained
image

FTR48 Version - Discretionary: maintained
image

FTR48 Version - Mandatory: no longer present
image

### DIFF
--- a/integration_tests/integration/recommendations/partA/recallType.spec.ts
+++ b/integration_tests/integration/recommendations/partA/recallType.spec.ts
@@ -15,6 +15,38 @@ context('Recall Type Page', () => {
     cy.signIn()
   })
 
+  const testRecallTypeRadioButton = (
+    radioElement: () => Cypress.Chainable<JQuery<Element>>,
+    radioGroup: () => Cypress.Chainable<JQuery<HTMLElement>>,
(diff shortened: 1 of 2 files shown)


ppmg:
Oh, I also forgot:
1. Can you include the ticket ID in the PR title so it's part of the eventual merge commit title, please?
2. Not sure if this is possible/easy to do, but the branch name is referring to the wrong ticket: MRD-2833 is the ticket for making the suitability page always available from the task list.

KiraNamida:
> Oh, I also forgot:
> 
> 1. Can you include the ticket ID in the PR title so it's part of the eventual merge commit title, please?
> 2. Not sure if this is possible/easy to do, but the branch name is referring to the wrong ticket: MRD-2833 is the ticket for making the suitability page always available from the task list.

2: Ah, I was on the wrong branch when I committed/pushed. let me close this one and replicate in the write place as it's small whilst dealing with that typo.

KiraNamida:
Recreated as #1401 with the commits against the correct branch

PR #1402, opened 2025-08-18
Mrd 2831 upgrade dependencies

EmilyHazlehurst:


.circleci/config.yml
```diff
@@ -267,9 +267,7 @@ jobs:
       - run:
           working_directory: make-recall-decision-e2e-tests
           name: Update npm
-          # Temporarily pinning the node version to allow pipeline builds to run.
-          # Can move back to npm@latest after upgrading to v10
-          command: 'npm install -g npm@9'
+          command: 'npm install -g npm@10'
```
ppmg, line 270: The previous comment said we could move to `npm@latest` once upgrading to v10 - is that not wise anymore (or maybe it never was?)?

ppmg (approved):
LGTM

PR #1403, opened 2025-08-18
MRD-2788: Notification banner amends

EmilyHazlehurst:
- Override the heading's max width to 100%

### DIFF
--- a/server/views/components/notification-banner/_index.scss
+++ b/server/views/components/notification-banner/_index.scss
@@ -1,0 +1,5 @@
+.car-notification-banner {
+  .govuk-notification-banner__heading {
+    max-width: 100%;
+  }
+}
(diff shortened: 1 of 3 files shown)


server/views/components/notification-banner/_index.scss
```diff
@@ -0,0 +1,5 @@
+.car-notification-banner {
+  .govuk-notification-banner__heading {
```
ppmg, line 2: Why the `.govuk-notification-banner__heading` part? Is it not enough to apply it to `.car-notification-banner`?

EmilyHazlehurst: The notification banner is already 100% width, it's the heading that's got a limited width so I need to target that specifically to override it.

ppmg (approved):
LGTM

PR #1410, opened 2025-08-28
MRD-2828: FTR48 Feature Toggle

KiraNamida:
[MRD-2828](https://dsdmoj.atlassian.net/browse/MRD-2828)

Set environment feature flag override to allow setting for all users and releasing a feature:

Previously, "feature flags" as this application calls them where only available in development environments and where a tool for development only. Now we want them for releasing a feature to Production at a specified date and time.
To support this the `featureFlags` middleware has been expanded to look for any environment keys that follow the format of `FEATURE_{flagName}` and, if the value is datetime and it currently past that date and time, then feature is enabled regardless of another other settings
This allows development flags to still function in a development context until we hit a threshold where it becomes enabled globally. The development flags remain cut off from production environments so can never be enabled until it is set for the environment.

Dev is being set the August 1st just so Dev goes straight into this environment wide enabled as we test and prepare for the release whilst preprod and prod are set in readiness of the September 2nd release date.

[MRD-2828]: https://dsdmoj.atlassian.net/browse/MRD-2828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/middleware/featureFlags.test.ts
```diff
@@ -1,54 +1,195 @@
...
+        })
+
+        it('Overrides a default of false if flag is "1" in request query', () => {
+          readFeatureFlags(testFlag(false))(queryReq(true), resultRes, next)
+          expect(resultRes.locals.flags).toEqual({
+            testFlag: true,
+          })
+        })
+
+        describe('Returns the cokkie/query value regardless of other flag values', () => {
```
ppmg, line 115: I think I've lost track a bit here, not sure if I'm misunderstanding or if something you intended is missing: what do you mean here by "regardless of other flag values"? Set in `process.env`? Or somewhere else?

Also, there's a typo in cokkie.

server/utils/utils.ts
```diff
@@ -141,6 +141,21 @@ export function isDateTimeRangeCurrent(startDateTimeStr: string, endDateTimeStr:
   return startDateTime <= dateToday && endDateTime >= dateToday
 }
 
+export function isDateTimeAfterCurrent(afterDateTimeStr: string) {
+  if (!afterDateTimeStr) {
+    return false
+  }
+
+  const dateToday = new Date()
```
ppmg, line 149: I'm not too familiar with dates in Javascript, so better to ask: should we not be using `Date.UTC` here and in the next line? The helm yaml files are expressed in UTC (maybe they should have the Z character to make that explicit?).

ppmg (changes requested):

PR #1478, opened 2025-12-24
MRD-3014 Present PPUD offence comment to users

ppmg:
In addition to displaying existing offence comments when the user has selected
an existing sentence from PPUD, a redesign of various pages and the order in
which they are presented is implemented here. The main reason for the redesign
in the order is that the Match Index Offence page had been until now been
presented to the user in the Determinate flow before they were presented with
the PPUD sentences to select from. As part of the re-ordering the design team
also determined it would be better to ask the user explicitly whether they
wished to make changes to the offence and/or offence comment in the case where
they have selected a Determinate PPUD sentence. This is all reflected in this
set of changes.


integration_tests/integration/recommendations/ppcs/util.ts
```diff
@@ -0,0 +1,10 @@
+import searchMappedUserResponse from '../../../../api/responses/searchMappedUsers.json'
+import searchActiveUsersResponse from '../../../../api/responses/ppudSearchActiveUsers.json'
+
+export function setUpSessionForPpcs() {
+  cy.session('login', () => {
+    cy.task('searchMappedUsers', { statusCode: 200, response: searchMappedUserResponse })
+    cy.task('ppudSearchActiveUsers', { statusCode: 200, response: searchActiveUsersResponse })
+    cy.signIn({ roles: ['ROLE_MAKE_RECALL_DECISION_PPCS'] })
+  })
```
EmilyHazlehurst: 🙌

server/@types/make-recall-decision-api/models/PpudDetailsResponse.ts
```diff
@@ -36,6 +36,7 @@ export type PpudDetailsSentence = {
 
 export type PpudDetailsOffence = {
   indexOffence?: string,
+  offenceComment?: string,
```
EmilyHazlehurst: Same as with the API PR. We also have indexOffenceComment e.g. 

```
export type PpudUpdateOffenceRequest = {
  indexOffence: string,
  indexOffenceComment?: string,
  dateOfIndexOffence: string,
};
```

Should we make them all either offenceComment/indexOffenceComment for clarity.

server/views/components/text-list/template.njk
```diff
@@ -0,0 +1,19 @@
+{% from "components/heading/macro.njk" import heading %}
+
+<div class="govuk-text-list{{ " " + params.classes if params.classes else ""}}">
```
EmilyHazlehurst: What component is this in the govuk design system?

If this is something that's been made custom for this project can we namespace it to us rather than govuk so that's clear? `car-text-list` or similar.

server/views/components/form-submit-button/template.njk
```diff
@@ -1,4 +1,4 @@
-<div class='govuk-!-margin-top-6'>
+<div class='govuk-!-margin-top-{{ params.marginTop if params.marginTop else "6" }}'>
```
EmilyHazlehurst: There's a possibility this is necessary due to the current state of our components but I don't like that we allowing any number to be passed here. Button spacing should be consistent across the site.

Was there a reason this needed to be changed?

integration_tests/integration/recommendation.spec.ts
```diff
@@ -2506,57 +2506,6 @@ context('Make a recommendation', () => {
       cy.getText('mappaLevel').should('contain', 'Unknown')
     })
 
-    it('match index offence', () => {
```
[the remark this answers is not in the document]
EmilyHazlehurst: Nevermind, it's just been moved.

EmilyHazlehurst (changes requested):

server/views/components/text-list/template.njk
ppmg: I hadn't realised it isn't a govuk component. Will amend to have that name 👍

EmilyHazlehurst (dismissed):
LGTM 👍

fincardy-moj (approved):
LGTM!

PR #1561, opened 2026-03-30
MRD-3144 back link should be browser back when OOH recall

fincardy-moj:


server/controllers/recommendation/sentenceInformationController.ts
```diff
@@ -12,14 +12,24 @@ import { STATUSES } from '../../middleware/recommendationStatusCheck'
...
     recommendation,
+    statuses,
     urlInfo: { basePath },
     user: { token, userId },
     flags,
   } = res.locals
 
+  // Need to check if an OOH recall to prevent backlink from sending PO to the first task list
+  // Out of hours recalls should have the AP_ statuses
+  const isOutOfHoursRecall = !!(statuses as RecommendationStatusResponse[])
```
sureshmoj, line 23: tiny one: the post already has `isApRationaleRecorded` for this can we use the same name here ?

sureshmoj (approved):

server/controllers/recommendation/sentenceInformationController.ts
sureshmoj: agreed to do this in the future by extracting the bits into a helper function - so no need to change this in this PR

PR #1562, opened 2026-03-30
MRD-3208 update back link text 

fincardy-moj:
This PR was originally for ticket MRD-3144 but for various reasons it became unpinned from it. 

That's now been cloned and updated to ticket MRD-3208

### DIFF
--- a/server/views/pages/recommendations/sentenceInformation.njk
+++ b/server/views/pages/recommendations/sentenceInformation.njk
@@ -7,7 +7,11 @@
   {% if pageData.backLinkUrl is not undefined %}
     {{ backLink({ text: 'Back to Consider a recall questions', href: pageData.backLinkUrl, doBrowserBack: false }) }}
   {% else %}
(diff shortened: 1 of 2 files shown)


server/views/pages/recommendations/sentenceInformation.njk
```diff
@@ -7,7 +7,11 @@
   {% if pageData.backLinkUrl is not undefined %}
     {{ backLink({ text: 'Back to Consider a recall questions', href: pageData.backLinkUrl, doBrowserBack: false }) }}
   {% else %}
-    {{ backLink() }}
+    {% if pageData.isOutOfHoursRecall %}
```
fincardy-moj, line 10: Nunjucks just doesn't have an `else if` option it seems...

sureshmoj: it has `elif` rather than `else if`

sureshmoj (approved):

PR #1566, opened 2026-04-08
MRD-3167: Added unit test assertion for earlier fix

sureshmoj:


server/controllers/recommendations/helpers/taskCompleteness.test.ts
```diff
@@ -317,7 +317,7 @@ describe('taskCompleteness', () => {
           recallType: { selected: { value: 'NO_RECALL' } },
           personOnProbation: {
             ...recommendationResponse.personOnProbation,
-            ftr56MappaReviewed: true,
+            ...(group === SentenceGroup.ADULT_SDS ? { ftr56MappaReviewed: true } : { ftr56MappaReviewed: undefined }),
```
sureshmoj, line 320: Without the earlier fix, this check would fail
this check adds ftr56MappaReviewed only for adult_sds and its not needed for youth_sds or other types of sentences

ppmg (approved):
LGTM

PR #1567, opened 2026-04-09
MRD-3166 Update README & docs

ppmg:


docs/e2e-tests.md
```diff
@@ -1,17 +1,19 @@
 # E2E tests
 
-The E2E tests are colocated in this repo, in the `e2e-tests` directory. They run using Cypress, and share commands with the integration tests.
+The E2E tests are in [the e2e tests repo](https://github.com/ministryofjustice/make-recall-decision-e2e-tests/). They
+run using Cypress.
 
 There's a HMPPS dev 'community of practice' talk on [how the E2E tests are set up](https://drive.google.com/file/d/1OeekvkViazrYNJXGMZrlM8UZU-Z71x6X/view).
```
fincardy-moj: This google link doesn't work for anyone who turned up after the migration. (I've always wanted to actually watch this video)

docs/setup-running.md
```diff
@@ -1,21 +1,25 @@
 # Running the app locally
 
+The app can be run locally in two ways: entirely locally, with other services our API service depends on mocked out, or
+connecting to the relevant services in dev. Both approaches have some setup in common.
+
 ## Setup
 
-Use the version of Node specified in the package.json 'engines' property.
+Use the version of Node specified in the package.json 'engines' property of the UI repo.
```
fincardy-moj: Teecccchnically, this states `^20` and implies we can run with any version after 20, might be better to point to the `.nvmrc` instead.

Admittedly, this is all this is a bit ad-hoc in the node ecosystem, but `.nvmrc` should be the way to go as it hooks into `nvm` itself:
https://www.nvmnode.com/extend/nvmrc.html#how-to-use-the-nvmrc-file

README.md
```diff
@@ -87,4 +87,4 @@ These should be ordered by priority, for example if there's a different banner f
 * Developer dashboard (AppInsights)
     * [Prod](https://portal.azure.com/#@nomsdigitechoutlook.onmicrosoft.com/dashboard/arm/subscriptions/a5ddf257-3b21-4ba9-a28c-ab30f751b383/resourcegroups/dashboards/providers/microsoft.portal/dashboards/c920c355-b321-4048-8795-230b9c5a2728)
```
fincardy-moj, line 88: Is it worth mentioning here that you need Security Clearance to view the prod ones of these? (I think that's required..)

fincardy-moj (changes requested):

docs/e2e-tests.md
fincardy-moj: Sorry, just noticed this was updated in the E2E repo (and I'll give that video a watch!), so will need to be updated here too.

README.md
ppmg: I've stated it for both. I'm not sure if the system requires it for preprod, but my understanding is that both preprod and prod access should only happen with SC clearance.

fincardy-moj (approved):
LGTM!

PR #1570, opened 2026-04-10
MRD-3093: Remove IntegratedOffenderManagement

sureshmoj:


integration_tests/integration/recommendationTaskList.spec.ts
```diff
@@ -87,7 +87,6 @@ context('Recommendation - task list', () => {
     cy.getElement('Add more details about vulnerabilities or needs Completed').should('exist')
     cy.getElement('Are there any victims in the victim contact scheme? Completed').should('exist')
     cy.getElement('Is Jane Bloggs in custody now? Completed').should('exist')
-    cy.getElement('Is Jane Bloggs under Integrated Offender Management (IOM)? Completed').should('exist')
```
ppmg, line 90: Is it not easier to remove the whole test entirely (and maybe all other tests that are specifically for pre-FTR56 cases)? Or is the plan to do this with the last PR for this ticket?

ppmg (approved):

integration_tests/integration/recommendationTaskList.spec.ts
sureshmoj: I was focussing only on the pages that the ticket mentioned - but happy to remove all ftr56 disabled tests if its not too much , as part of last ticket

PR #1574, opened 2026-04-14
MRD-3175 upgrade to node v24

fincardy-moj:


tsconfig.json
```diff
@@ -1,17 +1,14 @@
 {
+  "extends": "@tsconfig/node24/tsconfig.json",
```
fincardy-moj, line 2: This has been brought in line with the typescript template

data/recommendations/nomisIndexOffenceGenerator.ts
```diff
@@ -17,12 +17,12 @@ export const NomisIndexGenerator: DataGenerator<NomisIndexOffence, NomisIndexOff
     }
     if (options?.selectedIndex && options.selectedIndex <= offences.length) {
       return {
-        selected: offences.at(options.selectedIndex).offenderChargeId,
+        selected: offences?.[options.selectedIndex].offenderChargeId,
```
fincardy-moj, line 20: Updated the typescript version, ended up with an error using `.at` which is seems like only Dan used, so I've just updated to use the "proper" JavaScript method of accessing an index.

Dockerfile
```diff
@@ -1,35 +1,32 @@
 # Stage: base image
-FROM node:20.17-bookworm-slim as base
+FROM ghcr.io/ministryofjustice/hmpps-node:24-alpine AS base
```
fincardy-moj, line 2: Switched up to this version of the base image, so updated the rest of the this file.

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   browser-tools: circleci/browser-tools@1.4.8
-  hmpps: ministryofjustice/hmpps@10
+  hmpps: ministryofjustice/hmpps@11
```
fincardy-moj, line 5: Bumped this orb, because why not?

package.json
```diff
@@ -149,13 +150,16 @@
     "superagent": "^10.0.0",
     "ts-loader": "^9.5.1",
     "url-value-parser": "^2.2.0",
+    "util": "0.12.5",
+    "util.inherits": "1.0.3",
```
fincardy-moj, line 154: `assert`, `util` and `util.inherits` are basically polyfills for bunyan (our logging lib) on Node v24/Cypress 15 as Bunyan's quite old. I had a look and it still seems to be the preferred library, so dug out somewhere cypress 15 was being used alongside node 24 and bunyan which can be seen here:

https://github.com/ministryofjustice/hmpps-manage-people-on-probation-ui/commit/729f41c4c1c887028b25362eb3161796db5983c9

sureshmoj (dismissed):
LGTM 👍

.circleci/config.yml
```diff
@@ -282,7 +283,7 @@ jobs:
           name: Install curl
           command: 'apt-get update && apt-get install curl -y'
       - node/install:
-          node-version: '20' # node/install doesn't have a working_directory parameter, but the .nvmrc file is in a
+          node-version: '24' # node/install doesn't have a working_directory parameter, but the .nvmrc file is in a
```
ppmg, line 286: Should we be updating e2e repo accordingly for this?

integration_tests/cypress.config.ts
```diff
@@ -42,7 +42,7 @@ export default defineConfig({
     configFile: 'reporter-config.json',
   },
   taskTimeout: 60000,
-  numTestsKeptInMemory: 10,
+  // numTestsKeptInMemory: 10,
```
ppmg: Can you either remove it entirely or leave a comment explaining why it's commented out for now, please? If removing, maybe a reply to this review point explaining why it's being removed in case we ever need to come back and check?

ppmg (changes requested):

.circleci/config.yml
fincardy-moj: I was planning on upgrading the E2E repo after this, just wanted a successful deployment to dev first, so yep, that will happen!

integration_tests/cypress.config.ts
fincardy-moj: Whoops, meant to remove it completely once I'd checked it was working, will update!

ppmg (approved):
LGTM

fincardy-moj:
For posterity's sake: the `numTestsKeptInMemory` cypress config option was added an set to `10` in an earlier branch, but it turns out these are only useful to store in memory when running the Cypress UI version. As such, the defaults are `0` for the CLI version (`cypress run` command) and `50` for the UI version (`cypress open` command) - giving it an actual value will overwrite these defaults. 

Given our main use of this outside of development is the CLI version, it's best to keep this option out of the config and have the default of `0` kept in place for CLI.

PR #1577, opened 2026-04-16
MRD-3175 do full checkout for unit tests, define the missing env vars

fincardy-moj:


.circleci/config.yml
```diff
@@ -128,7 +128,8 @@ jobs:
       tag: << pipeline.parameters.node-version >>
     resource_class: xlarge
     steps:
-      - checkout
+      - checkout:
```
fincardy-moj, line 131: I *suspect* this is potentially related to upgrading the hmpps orb to v11 - though I can't for the life of me work out why. 

This error just rang a bell and I dug out this old thread of Pablo sorting it out like this: https://mojdt.slack.com/archives/C05MFCEV17E/p1764174775393589

.circleci/config.yml
```diff
@@ -347,6 +348,11 @@ jobs:
       - node/install:
           node-version: '24' # node/install doesn't have a working_directory parameter, but the .nvmrc file is in a
                              # sub-directory (make-recall-decision-e2e-tests), so we need to set the node version here
+      - run:
+          command: |
```
fincardy-moj, line 352: These were previously being hardcoded to nonsense values (and the `GIT_BRANCH` wasn't set at all), but they're actually being set, so we need to define them here so the script injects them into the Dockerfile correctly.

ppmg: I'm not sure I follow. This is in the context of local e2e tests, so we're not using the Dockerfile, but the docker-compose ones, aren't we?

fincardy-moj: Yeah, so the process here is that it calls the `./scripts/start-services-for-e2e-tests.sh -p` script, which in turn uses `docker-compose.yml` which does this:
```
  make-recall-decision-ui:
    build:
      context: .
 ```
 
 The `context: .` bit is saying "use the Dockerfile at the root of the repo" rather than specifying an image

ppmg (approved):
LGTM

PR #1579, opened 2026-04-16
MRD-3175 swap where we're defining build details

fincardy-moj:
It appears that it's not pulling through the environment variables and still throwing errors, so I've dug out where these are being set in the typescript template (which I'm trying to follow) and found this:

https://github.com/ministryofjustice/hmpps-template-typescript/pull/233/changes#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3

I suspect because this is a local test, it needs some defaults in there.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ parameters:
     default: false
   e2e-parallelism:
     type: integer
-    default: 6
+    default: 8
   e2e-parallelism-env:
     type: integer
     default: 1
@@ -154,8 +154,7 @@ jobs:
 
     steps:
       - checkout
-      - browser-tools/install_chrome:
-          use_cache: true
+      - browser-tools/install_chrome
       - browser-tools/install_chromedriver
       - node/install # this way we make sure we are aligned with the version used for the build
       - node/install-packages:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  browser-tools: circleci/browser-tools@1.4.8
+  browser-tools: circleci/browser-tools@2.4.1
```
fincardy-moj, line 4: Updated to a later version (not the latest as it's only been a few days) of browser-tools

.circleci/config.yml
```diff
@@ -154,9 +154,8 @@ jobs:
 
     steps:
       - checkout
-      - run: sudo apt-get update
-      - browser-tools/install-chrome
-      - browser-tools/install-chromedriver
+      - browser-tools/install_chrome
```
fincardy-moj, line 157: ... and they updated the command name for installing

.circleci/config.yml
```diff
@@ -154,9 +154,8 @@ jobs:
 
     steps:
       - checkout
-      - run: sudo apt-get update
```
fincardy-moj, line 157: Removed the `apt-get update` run here too, as it was causing everything to be slow and I couldn't find a reason for it to actually be there (the update only updates headers and package lists, so it's not actually installing anything)

sureshmoj (approved):
LGTM 👍

PR #1588, opened 2026-04-24
MRD-2716: Filtering Determinate sentences

bhanugnosis:
MRD-2716: Filtering Determinate sentences


integration_tests/integration/recommendations/ppcs/determinate/determinatePpudSentences.spec.ts
ppmg, line 1: Sorry, my fault for not noticing this when we we're looking at the code in the call earlier on. This test is for a different page altogether. You can undo the changes to this file.

The page you're updating [is routed to](https://github.com/ministryofjustice/make-recall-decision-ui/blob/main/server/routes/recommendations.ts#L529-L530) under the [select-ppud-sentence url](https://github.com/ministryofjustice/make-recall-decision-ui/blob/main/server/routes/paths/ppcs.ts#L16).

[This spec test](https://github.com/ministryofjustice/make-recall-decision-ui/blob/main/integration_tests/integration/recommendation.spec.ts#L2683) and the one immediately after it test that page. That spec file is very large, as it tests loads of pages (and does so poorly). We have been splitting tests out from it into page-specific spec files whenever we touched on a page. If you'd like the challenge, feel free to do so. I think most of the tests in [this folder](https://github.com/ministryofjustice/make-recall-decision-ui/tree/main/integration_tests/integration/recommendations/partA) are quite recent, so they should be good to see as an example of how we write these types of tests at the moment. Your test should be called `selectPpudSentence.spec.ts` and should go in [this folder](https://github.com/ministryofjustice/make-recall-decision-ui/tree/main/integration_tests/integration/recommendations/ppcs/determinate). The tests in that folder and the sub-folder within are probably good to see as examples too - I simply pointed at the other ones because they're more recent.

server/controllers/recommendation/selectPpudSentenceController.test.ts
```diff
@@ -1,57 +1,147 @@
...
+      expect(res.locals.offence).toBeDefined()
+      expect(res.locals.offence).toEqual(recommendation.nomisIndexOffence.allOptions[0])
+
+      expect(res.render).toHaveBeenCalledWith('pages/recommendations/selectPpudSentence')
+      expect(next).toHaveBeenCalled()
     })
   })
 
   describe('post', () => {
     describe('Valid data', () => {
```
ppmg, line 54: Why did you remove the comment that was here?

server/controllers/recommendation/selectPpudSentenceController.test.ts
```diff
@@ -1,57 +1,147 @@
...
+
+      expect(res.locals.offence).toBeDefined()
+      expect(res.locals.offence).toEqual(recommendation.nomisIndexOffence.allOptions[0])
+
+      expect(res.render).toHaveBeenCalledWith('pages/recommendations/selectPpudSentence')
+      expect(next).toHaveBeenCalled()
     })
   })
 
   describe('post', () => {
```
ppmg, line 53: Why are you changing the `post` tests? You're only changing the `get` function, aren't you?

server/controllers/recommendation/selectPpudSentenceController.test.ts
```diff
@@ -1,57 +1,147 @@
...
 import ppcsPaths from '../../routes/paths/ppcs'
+import { PpudDetailsSentence } from '../../@types/make-recall-decision-api/models/PpudDetailsResponse'
+import { ppudDetailsSentence } from '../../@types/make-recall-decision-api/models/ppud/PpudDetailsResponse.testFactory'
+import { getDeterminateSentences } from '../../helpers/ppudSentence/ppudSentenceHelper'
 
 jest.mock('../../data/makeDecisionApiClient')
+jest.mock('../../helpers/ppudSentence/ppudSentenceHelper')
 
 describe('Select Determinate PPUD Sentence Controller', () => {
   describe('get', () => {
```
ppmg, line 15: Was there something wrong with the existing test? The way it's set up makes it easy to understand what exactly goes wrong on a failure and it looks like it can be updated to include the function call mock and the respective checks without having to touch that many lines, but maybe I'm missing something?

ppmg (changes requested):

server/controllers/recommendation/selectPpudSentenceController.test.ts
bhanugnosis: Removed the mocking

integration_tests/integration/recommendations/ppcs/determinate/selectPpudSentence.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    cy.task('getRecommendation', { statusCode: 200, response: recommendation })
+  })
+
+  describe('Filtering', () => {
+    it('should only show determinate sentences and not indeterminate ones', () => {
+      cy.visit(testPageUrl)
+
+      cy.contains('Add your booking to PPUD - John Doe')
+
+      cy.get('.govuk-radios__input').should('exist')
```
ppmg, line 98: You could use the [testRadioButtons function](https://github.com/ministryofjustice/make-recall-decision-ui/blob/main/integration_tests/componentTests/radioButtons.tests.ts#L91) here. You can find other tests doing so already, if you want to see an example.

integration_tests/integration/recommendations/ppcs/determinate/selectPpudSentence.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { fakerEN_GB as faker } from '@faker-js/faker'
+import { RecommendationResponseGenerator } from '../../../../data/recommendations/recommendationGenerator'
+import CUSTODY_GROUP from '../../../../server/@types/make-recall-decision-api/models/ppud/CustodyGroup'
+import RECOMMENDATION_STATUS from '../../../../server/middleware/recommendationStatus'
+import setUpSessionForPpcs from '../ppcs/util'
+
+context('Select PPUD Sentence', () => {
```
ppmg, line 8: For radio button pages we normally test three cases:
* Page displaying as expected when there is no previous selection.
* Page displaying as expected when there is a previous selection.
* Page displaying expected errors when submitting with no selection.

You can see an example of this in [recallTypeIndeterminate.spec.ts](https://github.com/ministryofjustice/make-recall-decision-ui/blob/main/integration_tests/integration/recommendations/partA/recallTypeIndeterminate.spec.ts#L76-L143) (see only the highlighted lines; the first part is an older test that is due to be removed).

server/controllers/recommendation/selectPpudSentenceController.test.ts
```diff
@@ -30,6 +37,21 @@ describe('Select Determinate PPUD Sentence Controller', () => {
             expect(res.locals.offence).toEqual(recommendation.nomisIndexOffence.allOptions[0])
           })
         })
+        describe('Determinate sentences', async () => {
+          const req = mockReq({
```
ppmg: I don't think you need to set up mocks, call the controller or check the `render` and `next` calls. See the controller is called in the `beforeEach` above and the `render` and `next` checks are done below. Within this describe, you should only need to verify `res.locals.determinateSentences` is as you expect it and that the function was called as expected.

ppmg (changes requested):

integration_tests/integration/recommendations/ppcs/determinate/selectPpudSentence.spec.ts
bhanugnosis: Added TODO: in the njk file, radio buttons implementation is different in this file.

integration_tests/integration/recommendations/ppcs/determinate/selectPpudSentence.spec.ts
bhanugnosis: Answered

server/controllers/recommendation/selectPpudSentenceController.test.ts
bhanugnosis: Answered

integration_tests/integration/recommendations/ppcs/determinate/selectPpudSentence.spec.ts
ppmg: You're still missing the cases for the first and last bullet points above. I think you'll need each one to call `cy.task('getRecommendation', ...)` with its own recommendation, as the one you have now has something preselected.

ppmg (changes requested):

integration_tests/integration/recommendations/ppcs/determinate/selectPpudSentence.spec.ts
bhanugnosis: Done

integration_tests/integration/recommendations/partA/selectPpudSentence.spec.ts
```diff
@@ -76,5 +118,33 @@ context('Select PPUD Sentence', () => {
...
+      })
+      cy.task('getRecommendation', { statusCode: 200, response: recommendationWithNoSelectedSentence })
+    })
+    it('Displays error message when no sentence is selected', () => {
+      cy.visit(testPageUrl)
+
+      cy.get('button.govuk-button').click()
+
+      testForErrorPageTitle()
+      // No Summary errors check required as this page is not using summary component.
```
ppmg: It's not the `summary` component that `testForErrorSummary` checks, but the `error-summary` one, so you should be able to use the function here. See `sentenceInformation.spec.ts` for an example of a test for a page with radio buttons using it.

integration_tests/integration/recommendations/partA/selectPpudSentence.spec.ts
```diff
@@ -42,6 +43,40 @@ context('Select PPUD Sentence', () => {
...
+  const recommendationWithNoSelectedSentence = RecommendationResponseGenerator.generate({
+    nomisIndexOffence: {
+      selectedIndex: 1,
+    },
+    id: recommendationId,
+    bookRecallToPpud: {
+      firstName: 'John',
+      lastName: 'Doe',
+      custodyGroup: CUSTODY_GROUP.DETERMINATE,
+      ppudSentenceId: 'NEW_PUD',
```
ppmg: Can you set the `ppudSentenceId` to null instead of this invalid value, please?

integration_tests/integration/recommendations/partA/selectPpudSentence.spec.ts
ppmg: Small point: you can get rid of this call to `cy.task('getRecommendation', ...)` here, as you're overriding it in every case anyway. Conversely, you can keep the `cy.task('getStatuses', ...)` call here and remove it from the individual cases below.

ppmg (changes requested):

integration_tests/integration/recommendations/partA/selectPpudSentence.spec.ts
ppmg: As agreed, further changes need to be made to the nunjucks file to fix other problems before the `testForErrorSummary` function can be used in this test. Left to be done in a separate ticket.

ppmg (approved):
LGTM

PR #1589, opened 2026-04-27
MRD-3042: Recall received date/time fields to reflect when PPCS received the PartA

sureshmoj:
Screenshot 2026-04-27 at 15 33 44
Screenshot 2026-04-27 at 15 33 56

### DIFF
--- a/integration_tests/integration/recommendation.spec.ts
+++ b/integration_tests/integration/recommendation.spec.ts
@@ -2279,59 +2279,6 @@ context('Make a recommendation', () => {
       cy.get('#check-booking-details-content-1').should('contain', 'NOMIS number').should('be.visible')
     })
 
-    it('check booking details shows "You must enter a date" and "You must enter a time" when receivedDateTime is null', () => {
-      cy.task('getRecommendation', {
-        statusCode: 200,
-        response: {
-          ...completeRecommendationResponse,
-          recallConsideredList: null,
-          bookRecallToPpud: {
-            receivedDateTime: null,
-          },
-          prisonOffender: {},
-          ppudOffender: {},
-        },
-      })
-      cy.task('getStatuses', {
-        statusCode: 200,
-        response: [{ name: RECOMMENDATION_STATUS.SENT_TO_PPCS, active: true }],
-      })
-      cy.task('updateRecommendation', { statusCode: 200, response: recommendationResponse })
-
-      cy.visit(`/recommendations/252523937/check-booking-details`)
-
-      cy.get('#check-booking-recall-information-list').should('contain', 'You must enter a date')
-      cy.get('#check-booking-recall-information-list').should('contain', 'You must enter a time')
-    })
-
-    it('check booking details shows the recall received date and time when receivedDateTimeUpdatedByPpcs is set', () => {
-      cy.task('getRecommendation', {
-        statusCode: 200,
-        response: {
-          ...completeRecommendationResponse,
-          recallConsideredList: null,
-          bookRecallToPpud: {
-            receivedDateTime: '2024-01-31T15:17:58Z',
-            receivedDateTimeUpdatedByPpcs: true,
-          },
-          prisonOffender: {},
-          ppudOffender: {},
-        },
-      })
-      cy.task('getStatuses', {
-        statusCode: 200,
-        response: [{ name: RECOMMENDATION_STATUS.SENT_TO_PPCS, active: true }],
-      })
-      cy.task('updateRecommendation', { statusCode: 200, response: recommendationResponse })
-
-      cy.visit(`/recommendations/252523937/check-booking-details`)
-
-      cy.get('#check-booking-recall-information-list').should('not.contain', 'You must enter a date')
-      cy.get('#check-booking-recall-information-list').should('not.contain', 'You must enter a time')
-      cy.get('#check-booking-recall-information-list').should('contain', '31 January 2024')
-      cy.get('#check-booking-recall-information-list').should('contain', '15:17')
-    })
-
     it('edit name', () => {
       cy.task('getRecommendation', {
         statusCode: 200,
--- a/integration_tests/integration/recommendation.spec.ts
+++ b/integration_tests/integration/recommendation.spec.ts
@@ -2279,6 +2279,59 @@ context('Make a recommendation', () => {
       cy.get('#check-booking-details-content-1').should('contain', 'NOMIS number').should('be.visible')
     })
 
+    it('check booking details shows "You must enter a date" and "You must enter a time" when receivedDateTime is null', () => {
+      cy.task('getRecommendation', {
+        statusCode: 200,
+        response: {
+          ...completeRecommendationResponse,
+          recallConsideredList: null,
+          bookRecallToPpud: {
+            receivedDateTime: null,
+          },
+          prisonOffender: {},
+          ppudOffender: {},
+        },
+      })
+      cy.task('getStatuses', {
+        statusCode: 200,
+        response: [{ name: RECOMMENDATION_STATUS.SENT_TO_PPCS, active: true }],
+      })
+      cy.task('updateRecommendation', { statusCode: 200, response: recommendationResponse })
+
+      cy.visit(`/recommendations/252523937/check-booking-details`)
+
+      cy.get('#check-booking-recall-information-list').should('contain', 'You must enter a date')
+      cy.get('#check-booking-recall-information-list').should('contain', 'You must enter a time')
+    })
+
+    it('check booking details shows the recall received date and time when receivedDateTimeUpdatedByPpcs is set', () => {
+      cy.task('getRecommendation', {
+        statusCode: 200,
+        response: {
+          ...completeRecommendationResponse,
+          recallConsideredList: null,
+          bookRecallToPpud: {
+            receivedDateTime: '2024-01-31T15:17:58Z',
+            receivedDateTimeUpdatedByPpcs: true,
+          },
+          prisonOffender: {},
+          ppudOffender: {},
+        },
+      })
+      cy.task('getStatuses', {
+        statusCode: 200,
+        response: [{ name: RECOMMENDATION_STATUS.SENT_TO_PPCS, active: true }],
+      })
+      cy.task('updateRecommendation', { statusCode: 200, response: recommendationResponse })
+
+      cy.visit(`/recommendations/252523937/check-booking-details`)
+
+      cy.get('#check-booking-recall-information-list').should('not.contain', 'You must enter a date')
+      cy.get('#check-booking-recall-information-list').should('not.contain', 'You must enter a time')
+      cy.get('#check-booking-recall-information-list').should('contain', '31 January 2024')
+      cy.get('#check-booking-recall-information-list').should('contain', '15:17')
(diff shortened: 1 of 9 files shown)


integration_tests/integration/recommendation.spec.ts
```diff
@@ -2279,6 +2279,59 @@ context('Make a recommendation', () => {
       cy.get('#check-booking-details-content-1').should('contain', 'NOMIS number').should('be.visible')
     })
 
+    it('check booking details shows "You must enter a date" and "You must enter a time" when receivedDateTime is null', () => {
```
ppmg: Can you move this into a new `checkBookingDetails.spec.ts` file under the `ppcs` folder, please? We should try to avoid adding more to this file and instead break stuff out from it as we write tests for the individual pages.

integration_tests/integration/recommendation.spec.ts
```diff
@@ -2279,6 +2279,59 @@ context('Make a recommendation', () => {
...
+      cy.task('getRecommendation', {
+        statusCode: 200,
+        response: {
+          ...completeRecommendationResponse,
```
ppmg: Could you use the generator instead of the json, please? It's harder to find what is and isn't set in it and what's relevant in each case.

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
```diff
@@ -195,7 +195,7 @@ describe('get', () => {
           lastName: convertedLastName,
           cro: PRISON_OFFENDER_TEMPLATE.identifiers[0].value,
           prisonNumber: PRISON_OFFENDER_TEMPLATE.bookingNo,
-          receivedDateTime: SENT_TO_PPCS_STATUS_TEMPLATE.created,
+          receivedDateTime: null,
```
ppmg, line 192: I think you can probably get rid of `SENT_TO_PPCS_STATUS_TEMPLATE` altogether now.

server/controllers/recommendation/ppcs/checkBookingDetailsController.ts
```diff
@@ -199,6 +199,12 @@ async function get(_: Request, res: Response, next: NextFunction) {
     hasLastKnownAddress = !checkIfAddressesAreEmpty(addresses)
   }
 
+  // force ppcs users to enter the date and time they received the recall, even if it is the same as the decision date/time
+  // receivedDateTimeUpdatedByPpcs will be set once they submitted the edit recall received form, so we can use it to determine if the date/time has been updated
+  if (!recommendation.bookRecallToPpud.receivedDateTimeUpdatedByPpcs) {
```
ppmg: Couldn't we always initialise it to null instead? See the `if (!hasValue(recommendation.bookRecallToPpud))` block further up in this file.

server/views/pages/recommendations/checkBookingDetails.njk
```diff
@@ -236,8 +236,8 @@
...
-                            { key: 'Recall received time', value: formatDateTimeFromIsoString({ isoDate: recommendation.bookRecallToPpud.receivedDateTime, timeOnly: true }), actions: [{ id: editIdPrefix+'recallreceivedtime', text: editText, href: 'edit-recall-received-date-and-time', visuallyHidden: 'recall received time' }] },
+                            { key: 'Recall received date', value: '<span id="receivedDateTime"></span>' + blankSafe(formatDateTimeFromIsoString({ isoDate: recommendation.bookRecallToPpud.receivedDateTime, dateOnly: true }), true, 'You must enter a date'), actions: [{ id: editIdPrefix+'receivedDateTime', text: editText, href: 'edit-recall-received-date-and-time', visuallyHidden: 'recall received date' }], error: errors.receivedDateTime },
+                            { key: 'Recall received time', value: blankSafe(formatDateTimeFromIsoString({ isoDate: recommendation.bookRecallToPpud.receivedDateTime, timeOnly: true }), true, 'You must enter a time'), actions: [{ id: editIdPrefix+'receivedDateTime', text: editText, href: 'edit-recall-received-date-and-time', visuallyHidden: 'recall received time' }], error: errors.missingReceivedDateTime },
```
ppmg, line 240: I noticed you're using `errors.missingReceivedDateTime` here but `errors.receivedDateTime` in the previous one, but the error checking done in the controller seems like it'll only work for the previous one. I don't see anything regarding errors in the ticket, has this been discussed with UCD (should we have two separate error messages at the top, one for date and one for time, instead of a single one? If a single one, which should the error link point to? etc)?

ppmg (changes requested):

integration_tests/integration/recommendation.spec.ts
sureshmoj: moved the two new tests into checkBookingDetails.spec.ts

integration_tests/integration/recommendation.spec.ts
sureshmoj: done.

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
sureshmoj: its still being used in STATUSES_TEMPLATE ?

server/controllers/recommendation/ppcs/checkBookingDetailsController.ts
sureshmoj: you mean without needing the newly added flag receivedDateTimeUpdatedByPpcs? but then it  always sets to null when the page is reloaded?

server/views/pages/recommendations/checkBookingDetails.njk
sureshmoj: we don't need two different error message here, only one is enough like "Enter recall received date and time" because they cant fill one without the other as the `edit` link takes them to same page to full both date and time

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
ppmg: But is it still needed there?

server/controllers/recommendation/ppcs/checkBookingDetailsController.ts
ppmg: Only if `!hasValue(recommendation.bookRecallToPpud)`, i.e. it only does it the first time the page is loaded.

server/views/pages/recommendations/checkBookingDetails.njk
ppmg: I agree that either both or neither will be set, but I think how we display that is a UCD call. I'm not sure if there's already a case like this in CaR, so not sure how it should be shown in case of error.

integration_tests/integration/recommendations/ppcs/checkBookingDetails.spec.ts
```diff
@@ -0,0 +1,63 @@
+import RECOMMENDATION_STATUS from '../../../../server/middleware/recommendationStatus'
+import searchMappedUserResponse from '../../../../api/responses/searchMappedUsers.json'
+import searchActiveUsersResponse from '../../../../api/responses/ppudSearchActiveUsers.json'
+import { RecommendationResponseGenerator } from '../../../../data/recommendations/recommendationGenerator'
+
+context('Check Booking Details page', () => {
+  const recommendationResponse = RecommendationResponseGenerator.generate()
+
+  beforeEach(() => {
```
ppmg, line 8: Sorry, forgot about this until a call earlier on with Bhanu: there's a `setUpSessionForPpcs` function you can do instead,

ppmg (changes requested):

integration_tests/integration/recommendations/ppcs/checkBookingDetails.spec.ts
sureshmoj: fixed

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
sureshmoj: after latest changes, its still needed we set the field to null incase of non OOH recalls
`receivedDateTime: isOutOfHoursRecall ? decisionDateTime : null,`
Incase of OOH, we still set it to be `decisionDateTime` -  which I will confirm it with UCD

server/views/pages/recommendations/checkBookingDetails.njk
sureshmoj: as per this [chat](https://mojdt.slack.com/archives/C076M6JQ0P3/p1777448966281809), we want to keep the single error message like I implemented in the PR, UCD is going to come up with a new page to collapse two rows into single one and its a new ticket
Also incase of OOH its the current behaviour, the `receivedDateTime` is same as `decisionDateTime`

server/controllers/recommendation/ppcs/checkBookingDetailsController.ts
sureshmoj: yup, thanks for highlighting this, i double checked and fixed the logic as per above suggestion

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
ppmg: Now that we've confirmed we use `decisionDateTime` for OOH we can see `SENT_TO_PPCS_STATUS_TEMPLATE` isn't needed anymore in the test, right?

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
```diff
@@ -462,6 +462,35 @@ describe('get', () => {
       }),
     )
   })
+
+  it('clears the recall receivedDateTime when loading a non OOH recall', async () => {
```
ppmg: Is this not covered now by the `'load'` test you've updated above?

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
sureshmoj: true, I am not using `SENT_TO_PPCS_STATUS_TEMPLATE` in the above test or am i missing something ?

ppmg (changes requested):

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
ppmg: I just mean that I think you can remove it altogether from the test, since it isn't relevant to the OOH (from your previous comment I understood you were saying it might need to be kept in the test for that)

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
sureshmoj: agreed, removed it

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
sureshmoj: removed SENT_TO_PPCS_STATUS_TEMPLATE from the test class

ppmg (approved):
LGTM

PR #1590, opened 2026-04-29
MRD-3024 Integrate flipt feature flagging

fincardy-moj:
ITT: We attempt to implement Flipt.

The approach I've taken here is just to take the original list of feature flags and replace it with what's coming from Flipt. This allows us to keep the cookie/query/.env overrides whilst also being able to manage things in a more "central" location ie. Flipt becomes our source of truth for the existence of feature flags.

The other, fairly key, implementation detail here is that Flipt doesn't allow us to "separate out" flags within our namespace, that's to separate between API and UI flags which *may* want to differ, so this service now assumes that all UI flags are prefixed with `ui-` and then when that arrives in the UI project the `ui-` bit is removed. So for the FTR56 flag (currently `flagFTR56Enabled`), in flipt that would be `ui-flagFTR56Enabled` which would exist in `res.locals.flags` as `flagFTR56Enabled`


helm_deploy/values-dev.yaml
```diff
@@ -18,6 +18,8 @@ generic-service:
     COMPONENT_API_URL: "https://probation-frontend-components-dev.hmpps.service.justice.gov.uk"
     SENTRY_ENVIRONMENT: "DEV"
     ENVIRONMENT: "dev"
+    FLIPT_SERVER_URL: "https://feature-toggles-dev.hmpps.service.justice.gov.uk/"
```
fincardy-moj, line 21: We only use a namespace and a URL, no authentication is required - see the note on this section of the feature flag repo:
https://github.com/ministryofjustice/hmpps-feature-flags#setup

integration_tests/mockApis/flipt.ts
```diff
@@ -0,0 +1,23 @@
+import { stubFor } from './wiremock'
+import flags from '../support/featureFlagDefinitions'
```
fincardy-moj, line 2: These are just the flags for the flipt client mock, figured here we'd be better off just having a list in a separate file so it's a bit easier to find/update.

integration_tests/support/commands.ts
```diff
@@ -14,6 +14,7 @@ import getRecommendationsResponse from '../../api/responses/get-case-recommendat
 Cypress.Commands.add('signIn', (opts = {}) => {
   cy.task('stubSignIn', opts)
   cy.task('stubAuthUser')
+  cy.task('getFlags')
```
fincardy-moj, line 17: This seems to be the consistent place to pop this for it run on all tests (as all tests require a sign in) I tried a few other places but nothing worked as consistently as here.

integration_tests/support/featureFlagDefinitions.ts
```diff
@@ -0,0 +1,27 @@
+export default [
+  {
```
fincardy-moj, line 2: This file will basically contain a slightly stripped down duplicate of the flags on flipt, this ensures we can just consistently use a flag and still get access to it in tests with the cookie method we're using `cy.visit('/pageUrl?flagName=1')`

server/controllers/featureFlags/index.ts
```diff
@@ -2,17 +2,23 @@ import { Request, Response } from 'express'
...
-      description: featureFlagsDefaults[key].description,
-      default: featureFlagsDefaults[key].default,
-      override: globalyEnabled || res.locals.flags[key],
-      globalyEnabled,
-    }
-  })
+  const featureFlags = await featureFlagsDefaults()
+
+  res.locals.flagsList = Object.keys(res.locals.flags)
+    .sort((a, b) => a.localeCompare(b))
```
fincardy-moj, line 8: Gotta sort here because Object.keys doesn't guarantee order and the page started jumping all over the place on refresh

server/views/pages/featureFlags.njk
```diff
@@ -1,55 +1,75 @@
...
+                {
+                  html: flagActions,
+                  href: "#"
+                }
+              ]
+            }
+          }), summaryRows) %}
+        {% endfor %}
+
+        {{ govukSummaryList({ id: 'flagsSummaryList', rows: summaryRows }) }}
```
fincardy-moj, line 59: The existing radio options list looked very strange with the flipt flags as it's missing labels and sometimes descriptions, so I've done quick and dirty swap to a summaryList where it looks a bit more sensible.

.env.local.sample
```diff
@@ -17,5 +17,7 @@ CYPRESS_PASSWORD_SPO=
 CYPRESS_USERNAME_ACO=MAKE_RECALL_DECISION_ACO_USER
 CYPRESS_PASSWORD_ACO=
 COMPONENT_API_URL=https://probation-frontend-components-dev.hmpps.service.justice.gov.uk
+FLIPT_SERVER_URL=https://feature-toggles-dev.hmpps.service.justice.gov.uk/
```
fincardy-moj, line 20: Technnnically not needed as it'll just default to these values (except in production where it's marked as required), but figured it was worth adding them anyways.

server/services/featureFlagService.ts
```diff
@@ -0,0 +1,91 @@
...
+    } catch (error) {
+      logger.error(error, 'Error retrieving all flags')
+      return []
+    }
+  }
+
+  async isFeatureEnabled(key: string, flagType: FeatureFlagResponse['type']): Promise<boolean> {
+    try {
+      const evaluationArguments = {
+        entityId: this.user.username,
```
fincardy-moj, line 57: Whilst we're unlikely to ever use proper "segments" for A/B testing or anything, Flipt wants an entityId, so I figured I'd pipe in the proper username by default.

server/services/featureFlagService.ts
```diff
@@ -0,0 +1,91 @@
...
+        flagKey: key,
+        context: {
+          Time: new Date(Date.now()).toISOString(),
```
fincardy-moj, line 60: Sending this by default as it'll just ignore it on non-segmented flags, or use it for the date-based flags.

helm_deploy/values-dev.yaml
ppmg: The environments have the Flipt URL and namespace stored as secrets, so you should be able to do this once from the `values.yaml` file. See [the API PR for the Flipt changes](https://github.com/ministryofjustice/make-recall-decision-api/pull/1259/changes#diff-ec3bc7b12c40bb85b5af7769b24e122c148d2c0885807e8d497a5fd407fcc1f5) as an example.

server/controllers/featureFlags/index.ts
```diff
@@ -2,17 +2,23 @@ import { Request, Response } from 'express'
...
-      override: globalyEnabled || res.locals.flags[key],
-      globalyEnabled,
-    }
-  })
+  const featureFlags = await featureFlagsDefaults(res.locals.user)
+
+  res.locals.flagsList = Object.keys(res.locals.flags)
+    .sort((a, b) => a.localeCompare(b))
+    .map(key => {
+      const globalyEnabled = determineEnvFeatureOverride(key)
```
ppmg: Tiny thing (I know it was already like this before, but since you're updating the code...): should be globally, with a double L.

server/middleware/featureFlags.test.ts
```diff
@@ -76,37 +82,39 @@ describe('readFeatureFlags', () => {
...
+        it('Uses request query over cookies', async () => {
+          await readFeatureFlags()(cookieAndQueryReq(true, false), resultRes, next)
           expect(resultRes.locals.flags).toEqual({
             testFlag: false,
           })
         })
 
-        it('Overrides a default of false if flag is "1" in request query', () => {
-          readFeatureFlags(testFlag(false))(queryReq(true), resultRes, next)
+        it('Overrides a default of false if flag is "1" in request query', async () => {
```
ppmg, line 121: Sorry, I know you haven't changed the test name, but is the name inaccurate here or am I misreading something?

server/services/featureFlagService.ts
```diff
@@ -0,0 +1,91 @@
...
+        logger.error(err, 'Unable to connect to feature flag service')
+        throw Error('Unable to connect to feature flag service')
+      }
+    }
+
+    return this.client
+  }
+
+  async getAll(): Promise<FeatureFlagResponse[]> {
+    const cacheKey = 'allFlags'
```
ppmg: Doesn't the FliptClient already provide its own caching? That's my understanding of the polling interval and, as far as I understood, how the Java client class works. It polls every interval to get all the namespace data, including the flags and context-based logic attached to them, and updates its cache, so evaluations all actually happen in the client. The documentation isn't great on either the Java or JavaScript clients, though, so it's very possible I've misunderstood how it works (and I can see [this PR](https://github.com/ministryofjustice/manage-a-workforce-ui/pull/1230/changes#diff-04d687fd31870106c4a6ff331b2c3d85b3b029de730d8917cd57756ac1e63ee0) in another project did it with a cache).
I'm not quite sure what the purpose of the polling is otherwise, though, as for flags with context-based logic it can't know what context to provide at polling time.

server/views/pages/featureFlags.njk
```diff
@@ -1,55 +1,75 @@
...
+              classes: 'car-v-align-top',
+              text: flag.description if flag.description else
+                "(No Description)"
+            },
+            actions: {
+              classes: 'govuk-!-width-one-third',
+              items: [
+                {
+                  html: flagActions,
+                  href: "#"
```
ppmg, line 52: Not sure I understand this `href` value.

assets/scss/overrides/_local.scss
```diff
@@ -286,4 +286,8 @@ pre {
 .guidance-panel {
   border-left-color: govuk-colour('blue');
   background-color: #eaf4f9;
-}
\ No newline at end of file
+}
+
+.car-v-align-top {
```
ppmg, line 311: More curiosity than anything else: why was this needed?

ppmg (changes requested):

assets/scss/overrides/_local.scss
fincardy-moj: Because some of the data for the flags changed the `/flags` page looked a bit weird, so I switched it up to a govuk summary list. Unfortunately the middle column (denoted as "value") in a summary list uses a strict `vertical-align: bottom` CSS property and the "actions" section is usually just expecting links so it rarely presents as a problem. Using the radio options meant each row was taller than expected and the description was aligning to the bottom which made it look *really* weird.

So rather than reinventing the wheel, I just added in a quick utility class to fix that "value" column behaviour of aligning the to the bottom of the element. Figured it's an internal thing so it doesn't need to follow the govuk design system *too* closely. 

Hopefully that makes sense, can demonstrate if not!

helm_deploy/values-dev.yaml
fincardy-moj: Are these in there now? I can't seem to see them.

server/middleware/featureFlags.test.ts
fincardy-moj: I think this is correct, it's overriding the `false` default if the request query has the flag name in there set to `1` - so like `/url?flagFTR56Enabled=1` should override the default `false` of the `flagFTR56Enabled`.

Happy to discuss!

server/views/pages/featureFlags.njk
fincardy-moj: This is just a result of twisting the summary list a bit so it behaves the way we want. It's 99% fit for purpose, but is opinionated on a few things. By default anything in "actions" will have an `<a>` tag wrapped around it because it's expecting them to be links, so just putting a `#` in place will prevent that <a> tag from actually doing anything.

server/services/featureFlagService.ts
fincardy-moj: After some serious digging, yes! It does appear that there's a cache on the client side SDKs, so I'll get this cache removed.

helm_deploy/values-dev.yaml
ppmg: Oh, you're right, they're only in dev at the moment. See [this comment in the PPUD Automation Flipt ticket](https://dsdmoj.atlassian.net/browse/MRD-2909?focusedCommentId=669088) on how to add them. Could you add them to the preprod and prod environments, please?

server/middleware/featureFlags.test.ts
ppmg: But isn't this within the same `describe` as the three `it`s preceding it, for which the `beforeEach` sets the default flag value to `true`?

helm_deploy/values-dev.yaml
fincardy-moj: Note: After some discussion, it seems these comments were made based on an older implementation method of Flipt. The newer method **does not** require an API key, and it's safe to put the feature flag URLs into plain text environment variables as they're also documented in the public feature flag repo and therefore not a secret.

This was confirmed in Slack here: https://moj.enterprise.slack.com/archives/C09V5A412BZ/p1778851539274409

with the examples for ARN here:
https://github.com/search?q=repo%3Aministryofjustice%2Fhmpps-arns-assessment-platform-ui%20FEATURE_FLAG_URL&type=code

integration_tests/support/featureFlagDefinitions.ts
```diff
@@ -0,0 +1,27 @@
...
+  },
+  {
+    name: 'Delete Recommendation',
+    key: 'ui-flagDeleteRecommendation',
+    description:
+      'Development team use only - shows links on the Recommendations tab allowing any recommendation to be marked as deleted. Deleting a recommendation allows a new one to be created, if needed. The "deleted" recommendation will be retained in the database, and no data or audit info will be lost.',
+    enabled: false,
+  },
+  {
+    name: 'FTR56 Flag',
```
ppmg, line 22: Should this now be replaced by the newer FTR56SentenceConviction flag? I expected so, and thought some of the integration tests with the new flag would fail, but that doesn't seem to be the case. Not sure why, though, since the feature flag middleware bases its flag values on what comes from the Flipt server 🤔.

server/middleware/featureFlags.test.ts
ppmg: This is still pending.

server/middleware/featureFlags.test.ts
```diff
@@ -57,56 +58,67 @@ describe('readFeatureFlags', () => {
...
+
         expect(resultRes.locals.flags).toEqual({
           testFlag: def,
         })
       })
     })
   })
 
-  describe('When environment feature settings override is not enabled', () => {
+  describe('When FEATURE_FLAG_QUERY_PARAMETERS_ENABLED is enabled', () => {
```
fincardy-moj: There were a few things going on in here. 

Firstly: these tests were for when this setting (which enables/disables the ability to override feature flags) was originally *enabled* in this test (see ln87) so this describe block's name was wrong for a start.

server/middleware/featureFlags.test.ts
```diff
@@ -57,56 +58,67 @@ describe('readFeatureFlags', () => {
...
     invalidFeatureDateValueCases.forEach(({ name, value }) => {
       let featureDisabledProcessEnv: NodeJS.ProcessEnv
+
       beforeEach(() => {
-        process.env[featureKey(testFlagKey)] = `${value}`
-        process.env[flagQueriesEnabedKey] = `${true}`
+        jest.clearAllMocks()
+        process.env[featureKey('testFlag')] = `${value}`
+        process.env.FEATURE_FLAG_QUERY_PARAMETERS_ENABLED = `${true}`
```
fincardy-moj, line 89: This was pointlessly behind a variable (`flagQueriesEnabedKey`) so I've deleted that and its typo and made it more verbose and hopefully more obvious as to what it's doing.

server/middleware/featureFlags.test.ts
```diff
@@ -57,56 +58,67 @@ describe('readFeatureFlags', () => {
...
+
       afterEach(() => {
         process.env = INITAL_ENV
       })
+
       describe(`- Not enabled value: ${name} - ${value}`, () => {
-        it('Overrides a default of true if flag is "0" in request query', () => {
-          readFeatureFlags(testFlag(true))(queryReq(false), resultRes, next)
+        it('Overrides a default of true if flag is "0" in request query', async () => {
+          process.env[featureKey('testFlag')] = `${true}`
```
fincardy-moj: The value of the `testFlag` key is defined by the "invalidFeatureDateValueCases" which was purposefully never true, so this test was originally not really doing anything. This line actively sets it to true so we can confirm it's been flipped to false.

server/middleware/featureFlags.test.ts
```diff
@@ -115,23 +127,22 @@ describe('readFeatureFlags', () => {
...
+              jest.clearAllMocks()
               process.env.ENVIRONMENT = env
             })
             afterEach(() => {
               process.env = featureDisabledProcessEnv
             })
             describe(`- Environment value: ${env}`, () => {
               beforeEach(() => {
-                process.env[flagQueriesEnabedKey] = `${false}`
+                process.env.FEATURE_FLAG_QUERY_PARAMETERS_ENABLED = `${false}`
```
fincardy-moj, line 140: again, just more verbose (and typo free) with the name of the parameter here

.gitleaks/.gitleaksignore
```diff
@@ -1 +1,2 @@
 # The Fingerprint of false positive alerts can be added here to be excluded from future scans.
+integration_tests/support/featureFlagDefinitions.ts:generic-api-key:23
```
fincardy-moj, line 2: Updating the default flags mock started to show the new flag name as a false positive for a secret, but they're defined in a public repo so they're safe to expose.

PR #1591, opened 2026-05-01
Fix recommendationsList.spec.ts

ppmg:
The test was implicitly relying on other tests' wiremock set-up. Fixed here.


fincardy-moj (approved):
LGTM!

PR #1592, opened 2026-05-05
Mrd 3202 implement esbuild in UI

fincardy-moj:
This PR brings ESBuild from the ts template into our repo. 

First, this greatly reduces build time, locally it's running about about 5ish seconds to first start of app, rather than the 22ish seconds of the existing build process.

Probably a bit more importantly, it fixes a series of problems caused by the existing build script.

Here's some fun facts about our current build script:
- We serve unprocessed (and un-minified) JavaScript files straight from `/assets/js/` folder in the root of the project, but, confusingly, map them back to an endpoint just called `/assets`
- The application.css file is generated in the `/assets/stylesheets` folder which *should* just be processed and dumped into `/dist` but we don't do that..
- ..what we actually do is copy the `/assets` folder [in the Dockerfile ](https://github.com/ministryofjustice/make-recall-decision-ui/blob/13818187c13974f4bbf0938d083362d13d078d26/Dockerfile#L49) and [then just expose that to the `/assets/` endpoint](https://github.com/ministryofjustice/make-recall-decision-ui/blob/13818187c13974f4bbf0938d083362d13d078d26/server/middleware/setUpStaticResources.ts#L17) which allows anything dumped into that `assets` folder is then available for anyone to download.... [try it yourself!](https://consider-a-recall-dev.hmpps.service.justice.gov.uk/assets/Duncann-Cardours-PartA-SAMPLE.docx)
- There are some inconsistencies in the naming of the stylesheets, we're currently using `/assets/sass` for the folder name, and the "main" file (there isn't _technically_ a "main" file as it's just using anything in the sass folder as an entrypoint) is called `application.sass` but all the files being imported from there have the `.scss` extension. There **is** a distinction between `scss` and `sass` as they use a slightly different syntax, so this does make things a bit confusing when dev'ving and there's no good reason for it.


assets/js/govukFrontendInit.js
```diff
@@ -1,3 +1,3 @@
-import * as GOVUKFrontend from '../../../../../../assets/govuk/govuk-frontend.min.js'
+import * as GOVUKFrontend from 'govuk-frontend'
```
fincardy-moj, line 1: Import paths are fixed! So this weird thing just looks normal now :)

assets/js/vStyleDropdownArrow.js
```diff
@@ -19,3 +19,7 @@ function vStyleDropdownArrow(config) {
     `z"/></svg>`
   )
 }
+
+if (typeof window !== 'undefined') {
```
fincardy-moj, line 23: Originally this file was just being imported wholesale (no processing at all) so it was just creating a function called `vStyleDropdownArrow` in the global namespace and it was able to find that. ESBuild does things "correctly" in that it'll try and prevent global namespace pollution by enclosing functions like this and keeping them modular.

As a result, we need to reattach this function to the window.

assets/scss/app.scss
```diff
@@ -0,0 +1,22 @@
+$govuk-global-styles: true;
```
fincardy-moj, line 1: The file formerly known as `application.sass`, so here we're doing the following:
- make it an `scss` file for consistency
- update the syntax from `sass` to `scss` (namely semi-colons)
- remove all the different app imports and pop them in their own file called `overrides/local.scss` which is the preferred method
- make this the only file in this folder, so it becomes the entry point for compiling scss files

assets/scss/overrides/_local.scss
```diff
@@ -1,3 +1,23 @@
+@import "risk/predictor-scores";
```
fincardy-moj, line 1: This is where our custom (S)CSS files should be imported now. The TS template is using `_local` to indicate it's not a "directly processed" file (ie. it's imported elsewhere rather than esbuild being pointed towards it)

esbuild/app.config.js
```diff
@@ -0,0 +1,29 @@
...
+ * Build typescript application into CommonJS
+ */
+const getAppConfig = buildConfig => ({
+  entryPoints: globSync(buildConfig.app.entryPoints),
+  outdir: buildConfig.app.outDir,
+  bundle: false,
+  sourcemap: true,
+  platform: 'node',
+  format: 'cjs',
+  loader: {
```
fincardy-moj, line 16: This is the only place we deviate from the typescript template in this file, the template isn't watching nunjucks files for build as they're not marked as an entry point (they don't need to be compiled, it's just the nunjucks/express plugin processing them in real time) but we want the app to rebuild when a change is made. 

This code basically just removes anything attempting to actually parse them (ie. the "loader" for these files) now they're marked as an entry point (which happens in a different file in a mo..)
 
Stole this from here:
https://github.com/ministryofjustice/hmpps-arns-assessment-platform-ui/pull/173

esbuild/build.config.js
```diff
@@ -0,0 +1,52 @@
...
+    isProduction,
+    isWatchMode,
+
+    app: {
+      outDir: path.join(cwd, 'dist'),
+      entryPoints: globSync([
+        path.join(cwd, '*.ts'),
+        path.join(cwd, 'server/**/*.ts'),
+        path.join(cwd, 'api/**/*.ts'),
+        'server/**/*.njk',
```
fincardy-moj, line 22: This adds the nunjucks files as an entrypoint (see above), code stolen from here:
https://github.com/ministryofjustice/hmpps-arns-assessment-platform-ui/pull/173

server/middleware/setUpStaticResources.ts
```diff
@@ -11,11 +11,11 @@ export default function setUpStaticResources(): Router {
   router.use(compression())
 
   //  Static Resources Configuration
-  const cacheControl = { maxAge: config.staticResourceCacheDuration * 1000 }
+  const cacheControl = { maxAge: config.staticResourceCacheDuration }
 
   Array.of(
+    '/dist/assets',
     '/assets',
-    '/assets/stylesheets',
```
fincardy-moj, line 18: No.

server/utils/nunjucksSetup.ts
```diff
@@ -40,6 +43,16 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   app.locals.applicationInsightsConnectionString = config.applicationInsights.connectionString
   app.locals.applicationInsightsRoleName = defaultName()
 
+  // Read the asset manifest to find the built versions of our assets
+  let assetManifest: Record<string, string> = {}
+
+  try {
```
fincardy-moj, line 49: Now these files are being processed, we're cache-busting by adding the build hash at the end of the built filenames. So, for example, `application.css` becomes `application.[HASH].css` which would normally prevent the usual method of including via a `<link rel="style" />` tag as we don't know what the name of the file would be.

However, now when we build, we also build a `manifest.json` which lives in the `/assets` folder, this provides us a JS object which looks a bit like this:

```
{
  "/assets/css/app.css.map": "/assets/css/app.ETJFEEMK.css.map",
  "/assets/css/app.css": "/assets/css/app.ETJFEEMK.css",
   // ... other assets
}
```
As you can see, it's just an object which maps the names of the source files to the built files. 

We load this into memory here...

server/utils/nunjucksSetup.ts
```diff
@@ -109,4 +122,5 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('formatSentenceLength', formatSentenceLength)
   njkEnv.addGlobal('renderString', renderString)
   njkEnv.addFilter('isBefore', isBeforeDate)
+  njkEnv.addFilter('assetMap', (url: string) => assetManifest[url] || url)
```
fincardy-moj, line 125: ...and then add a nunjucks filter to find the hashed name of our files using the `assetMap` filter. 

So we can do something like:

```
<link rel="stylesheet" src="{{ '/assets/css/app.css' | assetMap }}" />
```

and it'll either resolve the correct hash name or just use the `/assets/css/app.css`

server/views/partials/layout.njk
```diff
@@ -55,18 +55,18 @@
 {% from "components/text-list/macro.njk" import textList %}
 
 {% block head %}
-  <link href="/assets/stylesheets/application.css?{{ version }}" rel="stylesheet"/>
+  <link href="{{ '/assets/css/app.css' | assetMap }}" rel="stylesheet"/>
   {% for css in feComponents.cssIncludes %}
     <link href="{{ css }}" nonce="{{ cspNonce }}" rel="stylesheet" />
   {% endfor %}
   <script src="/assets/applicationinsights-web.min.js"></script>
```
fincardy-moj, line 62: Certain scripts come from the node_modules folder, so don't require the `assetMap` filter as we're not processing them.

server/middleware/setUpStaticResources.ts
```diff
@@ -11,12 +11,10 @@ export default function setUpStaticResources(): Router {
...
 
   //  Static Resources Configuration
-  const cacheControl = { maxAge: config.staticResourceCacheDuration * 1000 }
+  const cacheControl = { maxAge: config.staticResourceCacheDuration }
 
   Array.of(
-    '/assets',
-    '/assets/stylesheets',
-    '/assets/js',
+    '/dist/assets',
```
fincardy-moj, line 17: Everything is copied to `/dist` now so we only need to serve from there.

server/config.ts
```diff
@@ -41,7 +41,7 @@ export interface PpudApiConfig extends ApiConfig {
 export default {
   https: production,
   applicationName: 'Consider a recall',
-  staticResourceCacheDuration: 20,
+  staticResourceCacheDuration: '1h',
```
fincardy-moj, line 44: 20 seconds seemed wholly unnecessary here, so went with the sensible default from the typescript template of 1hr.

Dockerfile
```diff
@@ -46,9 +46,6 @@ COPY --from=build --chown=appuser:appgroup \
 COPY --from=build --chown=appuser:appgroup \
         /app/build-info.json ./dist/build-info.json
 
-COPY --from=build --chown=appuser:appgroup \
```
fincardy-moj, line 49: Stop copying the entire `/assets` folder over to the Dockerfile, don't need it and it was exposing 2 `.docx` files which I can't see any reason to exist. 

They can be seen here and were added during the prototyping phase:
https://github.com/ministryofjustice/make-recall-decision-ui/tree/main/assets

package.json
```diff
@@ -6,19 +6,12 @@
   "license": "MIT",
   "scripts": {
     "prepare": "husky",
-    "copy-views": "cp -R server/views dist/server/",
```
fincardy-moj, line 9: A whole bunch of these commands were there just to support the build, so I've been able to remove a bunch of those.

bhanugnosis (approved):
Good work.

PR #1597, opened 2026-05-11
MRD-2780 Migrate to GitHub Actions

ppmg:
This set of changes migrates the repo away from CircleCI and onto GitHub
Actions.

Notable details of these changes include:
- The integration and e2e local tests are run in parallel jobs, for which
  cypress-split needed to be installed and configured (changes to package.json
  and cypress.config.ts). See the hmpps-assess-risks-and-needs-github-actions
  repo for details on how the splitting and parallelisation is done. This is
  also being done in the e2e repo.
- The contents of the start-services-for-e2e-tests.sh script have been
  simplified by copying over the contents of the start-services-integration.sh
  script that existed in the e2e test repo. This is so that the workflow in
  this and the e2e test and API pipelines are as similar as possible.
  The e2e script and the one in this repo were very similar, with this one
  accepting a couple of flag options. Those options aren't needed anymore (the
  auth images have multi-architecture support, removing the need for the -a
  flag, and we're OK with always pulling, removing the need for the -p flag).
- The snyk security scan has been added. This replaces the trivy scan, which
  had to be removed a few weeks back due to security issues in trivy. The snyk
  one has been added using the snyk-add.sh script in the hmpps-github-actions
  repo.
- The port used for Wiremock is changed here to 9091. This is to align it with
  the HMPPS node_integration_tests.yml workflow.
- The existing code was running a record-build-info script during the build
  step of the Dockerfile to record the build number and git commit hash into a
  build-info.json file, which was then read by the application to provide it in
  various places (e.g. health endpoint). This has been replaced with
  environment variables set during the deployment process, so this file is no
  longer needed. The consumers of the data have also been updated accordingly.
  This aligns us with other services and simplifies the build and deployment
  process.


bin/record-build-info
```diff
@@ -1,21 +0,0 @@
-#!/usr/bin/env node
```
fincardy-moj, line 1: Glad to see the back of this!

fincardy-moj (dismissed):

fincardy-moj (approved):
LGTM!

PR #1613, opened 2026-06-05
MRD-3227 fix UI CVEs

fincardy-moj:


server/middleware/setUpWebSession.ts
```diff
@@ -32,7 +32,7 @@ export default function setUpWebSession(): Router {
   router.use((req, res, next) => {
     const headerName = 'X-Request-Id'
     const oldValue = req.get(headerName)
-    const id = oldValue === undefined ? uuidv4() : oldValue
+    const id = oldValue === undefined ? randomUUID() : oldValue
```
fincardy-moj, line 35: UUID has been tricky in the last few months with vulnerabilities, so I figured "why not use the built-in node one?" 

After doing some searching, we _are_ using the built-in node one everywhere but here. So I've just updated it and removed the flaky dependency

ppmg (approved):
LGTM

PR #1615, opened 2026-06-08
Fix sentenceToCommitIndeterminate.spec.ts

ppmg:
The sentenceLength needs to be greater than zero. Otherwise, if set to zero,
the formatPpudSentenceLength function called later on returns nothing, but the
testSummaryList one requires a non-null value.


sureshmoj (approved):
LGTM 👍

PR #1617, opened 2026-06-09
Adds allowlist to ui

andrewl:
Adds the same allow-list that is in place in the preprod and prod environments


fincardy-moj (approved):
LGTM!

PR #1619, opened 2026-06-11
fix ui cves

fincardy-moj:
There was a vulnerability with a transitive dependency called `shell-quote` which was only used by a direct dependency called `concurrently` 

However, `concurrently` was only used in the old build process which has been superseded by ESBuild, so we just removed concurrently and we're back to secure :)


sureshmoj (approved):
LGTM 👍

PR #1624, opened 2026-06-17
MRD-3198: Make the CRO field mandatory in the PPCS

sureshmoj:


server/controllers/recommendation/editCroController.ts
```diff
@@ -16,7 +18,7 @@ async function get(_: Request, res: Response, next: NextFunction) {
     values: isDefined(errors)
       ? unsavedValues
       : {
-          cro,
+          cro: cro || recommendation.prisonOffender?.cro || '',
```
sureshmoj: This is for the ticket requirement: 
`Text area to be pre-populated with the CRO from the Part A as shown in the check booking details screen`

server/controllers/recommendation/editCroController.ts
```diff
@@ -35,6 +37,22 @@ async function post(req: Request, res: Response, _: NextFunction) {
 
   const recommendation = await getRecommendation(recommendationId, token)
 
+  const croFromPartA = recommendation.prisonOffender?.cro
+  const croFromPpud = recommendation.ppudOffender?.croOtherNumber
+  const croEmpty = !cro || cro.trim().length === 0
+
+  if (croEmpty && !croFromPartA && !croFromPpud) {
```
sureshmoj: empty submission of the form is fine as long as there is a cro from Part A or from PPUD, if both of them are mssing then empty submission should raise a Form `Enter the CRO` error

integration_tests/integration/recommendations/ppcs/editCro.spec.ts
ppmg, line 1: Small point: could you create a `checkBookingDetails` folder within this `ppcs` one and put this `.spec.ts` file in it, please?

integration_tests/integration/recommendations/ppcs/editCro.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      },
+    })
+    cy.task('getStatuses', {
+      statusCode: 200,
+      response: [{ name: RECOMMENDATION_STATUS.SENT_TO_PPCS, active: true }],
+    })
+
+    cy.visit(`/recommendations/${recommendationResponse.id}/edit-cro`)
+
+    cy.get('#cro').clear()
```
ppmg: Is this step needed? Shouldn't it already be clear given how you've set up the response? Similarly in other tests where the field is expected to be clear from the start.

integration_tests/integration/recommendations/ppcs/editCro.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    cy.get('button.govuk-button').click()
+
+    cy.get('.govuk-error-summary').should('contain', 'There is a problem')
```
ppmg: You can call the `testForErrorPageTitle` and `testForErrorSummary` utility functions in `errors.tests.ts` instead. Also, the pattern we've been following is to have a `describe` for the page data and one for errors. Could you align it with that pattern, please? See `matchIndexOffence.spec.ts` for an example.

server/views/pages/recommendations/addressDetails.njk
```diff
@@ -31,7 +31,7 @@
                 <div class='govuk-!-margin-bottom-6'>
                     {% if recommendation.personOnProbation.addresses.length %}
 
-                        <p class="govuk=body">{% if recommendation.personOnProbation.addresses.length === 1 %}This is the last known address{% else %}These are the last known addresses{% endif %}
+                        <p class="govuk-body">{% if recommendation.personOnProbation.addresses.length === 1 %}This is the last known address{% else %}These are the last known addresses{% endif %}
```
ppmg, line 34: Nice catch! Thanks for fixing it!

server/views/pages/recommendations/alreadyExisting.njk
```diff
@@ -11,7 +11,7 @@
 
         <h1 class='govuk-heading-l'>{{ pageHeadings[page.id] }}</h1>
         <div >
-            <p class="govuk=body">
+            <p class="govuk-body">
```
ppmg, line 14: Nice catch! Thanks for fixing it!

server/controllers/recommendation/editCroController.ts
```diff
@@ -35,6 +37,22 @@ async function post(req: Request, res: Response, _: NextFunction) {
 
   const recommendation = await getRecommendation(recommendationId, token)
 
+  const croFromPartA = recommendation.prisonOffender?.cro
```
ppmg: The Part A mapping [first tries to use the CRO found in NDelius, then falls back to NOMIS](https://github.com/ministryofjustice/make-recall-decision-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt#L194). We should do the same here.

server/controllers/recommendation/editCroController.ts
ppmg: The Part A mapping [first tries to use the CRO found in NDelius, then falls back to NOMIS](https://github.com/ministryofjustice/make-recall-decision-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt#L194). We should do the same here.

server/controllers/recommendation/editCroController.ts
ppmg: I'm not sure the ` && !croFromPartA && !croFromPpud` part should be here. From what I'm seeing in the ticket's tables describing the behaviour for each case, the error message should be shown for empty state validation only. There's no mention there of having Part A or PPUD CROs.

server/views/pages/recommendations/editCro.njk
```diff
@@ -13,32 +13,40 @@
...
-                    <h3 class='govuk-heading-m'>From PPUD</h3>
-                    <div class="car-text-list__row">
-                        <dt class="car-text-list__key">CRO</dt>
-                        <dd class="car-text-list__value" data-qa="ppudCro">{{ recommendation.ppudOffender.croOtherNumber }}</dd>
-                    </div>
-                </dl>
+            {% if recommendation.ppudOffender %}
+                <h2 class='govuk-heading-m govuk-!-margin-bottom-0'>From PPUD</h2>
+                {% if recommendation.ppudOffender.croOtherNumber %}
+                    <dl class="car-text-list">
```
ppmg: I think you can use the `textList` macro here, can't you?

server/views/pages/recommendations/editCro.njk
```diff
@@ -13,32 +13,40 @@
...
+                        <div class="car-text-list__row">
+                            <dt class="car-text-list__key">CRO</dt>
+                            <dd class="car-text-list__value" data-qa="ppudCro">{{ recommendation.ppudOffender.croOtherNumber }}</dd>
```
ppmg: Since we're already touching on these files, could you switch this file and the controller such that the controller provides these fields to the njk file through the `values` field instead of expecting the njk to know where in the recommendation to dig around, please? So something like a `ppudCro` field within the `values` one, so that the njk just has to do `values.ppudCro`.

Similarly for the other values used in this same file, please.

server/views/pages/recommendations/editCro.njk
```diff
@@ -13,32 +13,40 @@
 
             <h1 class="govuk-heading-l">{{ pageHeadings[page.id] }}</h1>
 
+            <p class="govuk-body">
```
ppmg: While we're changing this file, would you mind using the `bodyParagraph` here and the `heading` one above, please?

integration_tests/integration/recommendations/ppcs/editCro.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      },
+    })
+    cy.task('getStatuses', {
+      statusCode: 200,
+      response: [{ name: RECOMMENDATION_STATUS.SENT_TO_PPCS, active: true }],
+    })
+
+    cy.visit(`/recommendations/${recommendationResponse.id}/edit-cro`)
+
+    cy.get('[data-qa="nomisCro"]').should('contain', '789/01A')
```
ppmg: My understanding is that data QA IDs aren't best practice anymore, as they don't reflect how users interact with our service. Instead, we should be searching for the elements we expect to see on our screen and check they are displayed/styled as expected. E.g. from this test we don't know if the CRO number is displayed as a text field, a heading, a button or the header on a table and the test wouldn't fail if we accidentally changed how it's displayed.

Other more recent tests we've added to the codebase test pages without relying on data QA IDs (e.g. `matchIndexOffence.spec.ts`), see how they do it for an example.

server/controllers/recommendation/editCroController.test.ts
```diff
@@ -30,6 +30,48 @@ describe('get', () => {
     expect(res.render).toHaveBeenCalledWith('pages/recommendations/editCro')
     expect(next).toHaveBeenCalled()
   })
+
+  it('pre-populates cro from Part A when bookRecallToPpud cro is empty', async () => {
```
ppmg: This will need to be turned into two tests once the controller is updated to also consider the NDelius CRO (see comment in the controller).

ppmg (changes requested):

server/controllers/recommendation/editCroController.ts
sureshmoj: i thought the fallback already happens in the api before UI gets the rec' response from api, but i was wrong, thanks for flagging , fixing this

server/controllers/recommendation/editCroController.ts
sureshmoj: yeah, not sure why I got that feeling of empty submission should be allowed if there is either part A cro or ppud cro is present - but agree with you with simple empty form validation - fixing it now

integration_tests/integration/recommendations/ppcs/editCro.spec.ts
sureshmoj: fixed

integration_tests/integration/recommendations/ppcs/editCro.spec.ts
sureshmoj: fixed

server/controllers/recommendation/editCroController.ts
sureshmoj: this line is no longer needed, so removed it

server/views/pages/recommendations/editCro.njk
sureshmoj: fixed

server/views/pages/recommendations/editCro.njk
sureshmoj: fixed

server/views/pages/recommendations/editCro.njk
sureshmoj: fixed

integration_tests/integration/recommendations/ppcs/editCro.spec.ts
sureshmoj: fixed

server/controllers/recommendation/editCroController.test.ts
sureshmoj: That's already done - the current tests already cover both fallback paths separately:                                                                                                                            
                                                                                                                                                                                                                   
  1. 'pre-populates cro from personOnProbation.croNumber when bookRecallToPpud cro is empty' - tests the NDelius (personOnProbation) fallback first                                                                
  2. 'pre-populates cro from prisonOffender.cro when bookRecallToPpud and personOnProbation cro are empty' - tests the Part A (prisonOffender) fallback when NDelius is also empty

or am I missing something else?

integration_tests/integration/recommendation.spec.ts
ppmg: I think we can get rid of this test now that you've added yours, can't we?

server/controllers/recommendation/editCroController.test.ts
```diff
@@ -40,6 +43,33 @@ describe('get', () => {
...
+          ppudOffender: { croOtherNumber: '333/44B' },
+        },
+      },
+    })
+    const next = mockNext()
+    await editCroController.get(req, res, next)
+
+    expect(res.locals.values).toEqual({
+      cro: '111/22A',
+      partACro: '789/01A',
```
ppmg: This should also be the nDelius one. The NOMIS one is a fallback if nDelius doesn't have it.

server/controllers/recommendation/editCroController.test.ts
```diff
@@ -40,6 +43,33 @@ describe('get', () => {
...
+      ppudCro: '333/44B',
+      hasPpudRecord: true,
+    })
+  })
+
+  it('pre-populates cro from partACro when bookRecallToPpud and ndeliusCro are empty', async () => {
```
ppmg: Following on the previous comment, this should say nomisCro rather than partACro

server/controllers/recommendation/editCroController.ts
```diff
@@ -8,18 +8,23 @@ import strings from '../../textStrings/en'
 async function get(_: Request, res: Response, next: NextFunction) {
   const { recommendation, errors, unsavedValues } = res.locals
   const { cro } = recommendation.bookRecallToPpud
+  const ndeliusCro = recommendation.personOnProbation?.croNumber || null
+  const partACro = recommendation.prisonOffender?.cro || null
```
ppmg: Sorry, I meant that the Part A value should be nDelius || NOMIS || null (in that order), to align with what the back-end does.

That being said, I just realised [we currently pre-fill it just with the NOMIS value](https://github.com/ministryofjustice/make-recall-decision-ui/blob/main/server/controllers/recommendation/ppcs/checkBookingDetailsController.ts#L110). Should we just be changing that instead and leaving this editing page be purely about displaying the existing `bookRecallToPpud.cro` and editing it?

server/controllers/recommendation/editCroController.ts
```diff
@@ -8,18 +8,23 @@ import strings from '../../textStrings/en'
...
-    values: isDefined(errors)
-      ? unsavedValues
-      : {
-          cro: cro || recommendation.prisonOffender?.cro || '',
-        },
+    values: {
+      partACro,
+      ppudCro,
+      hasPpudRecord,
+      cro: isDefined(errors) ? unsavedValues?.cro : cro || ndeliusCro || partACro || '',
```
ppmg: Should we ever fall back to the PPUD value? There's no AC in the ticket saying so, but there also isn't an AC describing the case where the Part A has no CRO but PPUD does, so I'm wondering if that was a deliberate omission.

server/views/pages/recommendations/editCro.njk
ppmg: Small point: missed the heading above this line, the one with the heading for the page. I think the default parameters for the `heading` macro is designed for page headings, so is probably just a call to the macro with `pageHeadings[page.id]`.

server/controllers/recommendation/editCroController.test.ts
ppmg: That was what I meant, yep, although see my newer comment in the controller regarding whether this controller should be doing any of this logic.

ppmg (changes requested):

server/views/pages/recommendations/editCro.njk
sureshmoj: fixed

integration_tests/integration/recommendation.spec.ts
sureshmoj: fixed

server/controllers/recommendation/editCroController.ts
sureshmoj: got it, fixed, thanks 
As part of MRD-3209 we are making the CRO field as mandatory, so I guess as part of that we should show what ever user submits in this edirCro page  ?
Using just nomis value in checkBookingDetailsController looks like a bug though

server/controllers/recommendation/editCroController.ts
sureshmoj: not sure but I will ask UCD.

server/controllers/recommendation/editCroController.test.ts
sureshmoj: fixed

server/controllers/recommendation/editCroController.test.ts
sureshmoj: fixed

server/controllers/recommendation/editCroController.ts
ppmg: But users always go through the Check booking details page first, so `checkBookingDetailsController` is responsible for initialising the fields, not the editing pages. I agree that just using the NOMIS value in `checkBookingDetailsController` would be according to the new intended behaviour, but what I meant from the previous comment is that I think the behaviour you're implementing should actually be added to `checkBookingDetailsController`, not to `editCroController`. By the time the user arrives at the latter, they've already gone through the former.

It makes sense as well that it's `checkBookingDetailsController` is responsible for this. Were we to follow the pattern of each edit controller being responsible for the initialisation part of its field, then users would be forced to go into every edit page to trigger it, rather than it happening automatically once when the booking details are first loaded.

sureshmoj: fixed, 
fallback happens in checkBookingDetailsController now.

server/controllers/recommendation/editCroController.ts
```diff
@@ -9,7 +9,8 @@ async function get(_: Request, res: Response, next: NextFunction) {
   const { recommendation, errors, unsavedValues } = res.locals
   const { cro } = recommendation.bookRecallToPpud
   const ndeliusCro = recommendation.personOnProbation?.croNumber || null
-  const partACro = recommendation.prisonOffender?.cro || null
+  const nomisCro = recommendation.prisonOffender?.cro || null
+  const partACro = ndeliusCro || nomisCro || null
```
sureshmoj, line 13: we still need this `partACro` value so njk file uses it to show '`From Part A`' field in the UI

server/controllers/recommendation/editCroController.ts
sureshmoj: added a fallback to ppud cro in the latest commit

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
```diff
@@ -290,6 +435,7 @@ describe('get', () => {
             lastName: 'Bloggs',
             dateOfBirth: '2000-01-01',
             prisonNumber: '1234',
+            cro: '123X',
```
ppmg: Could you set this to be `RECOMMENDATION_TEMPLATE.personOnProbation.croNumber` so it's clear where this is coming from, please?

integration_tests/integration/recommendations/ppcs/checkBookingDetails/editCro.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+  })
+
+  describe('Page Data', () => {
+    it('pre-populates CRO input from bookRecallToPpud cro', () => {
+      cy.task('getRecommendation', {
+        statusCode: 200,
+        response: {
+          ...recommendationResponse,
+          bookRecallToPpud: { cro: '999/99A' },
+          personOnProbation: { ...recommendationResponse.personOnProbation, croNumber: '111/22A' },
```
ppmg: Could you remove any of the specific set-up here that isn't needed for the test case (the `personOnProbation`, `prisonOffender` and `ppudOffender` parts), please? So it's clear what is and isn't important for it. We can keep the `...recommendationResponse` part, so it shows we can have the other values but we don't care what they are here.

Similarly for the other cases below, please.

ppmg (changes requested):

integration_tests/integration/recommendations/ppcs/checkBookingDetails/editCro.spec.ts
sureshmoj: fixed

server/controllers/recommendation/ppcs/checkBookingDetailsController.test.ts
sureshmoj: fixed

ppmg (commented):
LGTM! Holding off on approval and merge for now until pipeline is a bit clearer

ppmg (approved):

PR #1632, opened 2026-06-25
Fix for UI CVEs

bhanugnosis:
Fix for UI CVEs: applicationinsights to 3.15.1 and jest & jest peer dependencies to 30.4.2


ppmg (approved):
LGTM

PR #1636, opened 2026-06-30
MRD-3188 Update security config for SmartSurvey pop-up

ppmg:
Custom styling and an image have been added to the pop-up, so we need to update
the security config to allow them through.


sureshmoj (approved):
LGTM 👍

